PR #12, opened 2021-07-30
Feature/transform app test 

pietrzakacper:
End-to-end test of our existing transform-app

### DIFF
--- a/packages/reference-apps/transform-sequence-1/package.json
+++ b/packages/reference-apps/transform-sequence-1/package.json
@@ -19,5 +19,12 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/scramjetorg/transform-hub.git"
+  },
+  "dependencies": {
+    "scramjet": "^4.35.12"
+  },
+  "publishConfig": {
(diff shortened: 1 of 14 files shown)


packages/reference-apps/transform-sequence-1/package.json
```diff
@@ -19,5 +19,12 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/scramjetorg/transform-hub.git"
+  },
+  "dependencies": {
+    "scramjet": "^4.35.12"
+  },
+  "publishConfig": {
```
stokowski, line 26: we are fully moving to npmjs, so this won't be necessary

stokowski (changes requested):

packages/reference-apps/transform-sequence-1/package.json
a-tylenda: `dependencies` and `publishConfig`  removed in commit [ed25b4](https://github.com/scramjetorg/transform-hub/pull/12/commits/ed25b4f6c52ae573ffd937f3e9010644648c04d4)

stokowski (approved):

PR #40, opened 2021-08-10
Fix CLI start command to allow passing variadic arguments in a civilized way

pietrzakacper:
Before we had to pass arguments to the start command like this:
```
si seq start 27cb6496-5895-40a9-9087-7a3b4c26253a '["BTC, "ETH"]'
```

now it's possible to do it the normal way:

```
si seq start 27cb6496-5895-40a9-9087-7a3b4c26253a BTC ETH
```

and the bonus is, we support passing no arguments ;)

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -7,7 +7,7 @@
     "si": "./src/bin/index.ts"
   },
   "scripts": {
-    "start": "ts-node ./src/index",
+    "start": "ts-node ./src/bin/index",
     "build": "tsc -p tsconfig.build.json",
     "clean": "rm -rf ./dist .bic_cache",
     "test": "echo no tests yet -- # npm run test:ava",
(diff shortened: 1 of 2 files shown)


packages/cli/package.json
```diff
@@ -7,7 +7,7 @@
     "si": "./src/bin/index.ts"
   },
   "scripts": {
-    "start": "ts-node ./src/index",
+    "start": "ts-node ./src/bin/index",
```
jan-warchol, line 10: why this change? it seems unrelated fo arg parsing.

jan-warchol (commented):
And that was it? Nice!

packages/cli/package.json
pietrzakacper: it's needed to run and test this package locally

pietrzakacper: ./src/index file does not exists ;)

MichalCz (approved):

packages/cli/package.json
jan-warchol: ah, so it was broken from the start? pity.

PR #49, opened 2021-08-17
Add changes in checksum test

a-tylenda:
asana task: https://app.asana.com/0/1200747048063058/1200791457586540


packages/reference-apps/checksum-sequence/index.ts
```diff
@@ -1,23 +1,36 @@
...
     this.logger.info("Sequence Checksum is called.");
-    let out = "";
 
-    for await (const chunk of _stream) {
-        this.logger.info("Sequence Checksum received chunk: " + chunk.toString());
-        out += decoder.write(chunk);
-    }
-    out += decoder.end();
-    const hex = crypto.createHash("md5").update(out).digest("hex");
+    const jsonToString = readFileSync(ffrom).toString();
```
alicja-gruzdz, line 11: In the Instance we should read from the _stream (input).
The idea was to check whether we can send all data without any packet lost to the Instance using input stream (/input endpoint). Let's talk tomorrow morning and look into it together :)

PR #50, opened 2021-08-17
Fix E2E-004 TC-002 test by modifying event-sequence-2 ref app

pietrzakacper:
Tbh I am not entirely sure how this is a fix.
I was getting this error in runtime of this sequence:
```
2021-08-17T16:57:15.666Z error (object:Runner) Error occured during sequence execution:  TypeError: o[Symbol.iterator] is not a function
    at __asyncValues (/package/index.js:26:98)
    at RunnerAppContext.<anonymous> (/package/index.js:56:33)
    at Generator.next (<anonymous>)
    at /package/index.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/package/index.js:4:12)
    at RunnerAppContext.exp (/package/index.js:50:16)
    at Runner.<anonymous> (/app/runner/runner.js:365:32)
    at Generator.next (<anonymous>)
    at fulfilled (/app/runner/runner.js:5:58)
```

I suspected that it might be some bug with the way async generators are polyfilled.

Tests pass now though ✅

https://app.asana.com/0/1200747048063058/1200787625956059

### DIFF
--- a/packages/reference-apps/event-sequence-2/index.ts
+++ b/packages/reference-apps/event-sequence-2/index.ts
@@ -7,7 +7,7 @@ const exp: [ReadableApp<{a: number}, [], {x: number}>, WritableApp<{a: number}, 
      * @param _stream - dummy input
      * @returns data
      */
-    async function(_stream) {
+    async function*(_stream) {
         const data = this.initialState;
 
         let x = data?.x || 0;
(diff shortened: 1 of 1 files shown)


MichalCz (commented):
Hmm... This seems strange. Looking at [`runner/src/runner.ts`](https://github.com/scramjetorg/transform-hub/blob/d57d5db5a1a1e35c2871fec804a2a558bf8b493f/packages/runner/src/runner.ts) around line 499 - this should not make any difference (a function returning an async generator and a function returning it should be turned into a DataStream).

Do we actually polyfill the async generator at all? We shouldn't since we target node 12+ which supports both. Can you check ts settings?

a-tylenda (approved):

Issue #39, opened 2021-08-10
Sequence status: 422 Unprocessable Entity for a environment without downloaded Docker images


luccam-scr:
Sequence send returns "sequence status: 422 Unprocessable Entity" error after sending new sequence to environment without Docker images downloaded.
What I did is that I cleared all sth docker images and volumes from my local Docker engine (Win10, Docker integratet with Ubuntu via WSLv2).
First, second ans third try to start sequence returns error as below, you need to wait 2-3 minutes (probably for Docker containers start)  and later sequence start works ok.

Error:
Request ok: http://127.0.0.1:8000/api/v1/sequence status: 422 Unprocessable Entity
Error: Invalid id: undefined
    at new SequenceClient

luccam-scr:
Temporary solution (before we release fix) is to wait 2-3 minutes for STH to download Docker images and try to execute "si sequence send <package>" again. 
Once Docker images are available locally,  "si sequence send <package>" should return properly "Request ok: http://127.0.0.1:8000/api/v1/sequence status: 202 Accepted"

Issue #32, opened 2021-08-05
Docker adapter should return error if docker is not running


luccam-scr:
It happened that docker daemon was switched off, despite that STH started (yarn start) and did not return any error. What is expected that adapter fails in this situation.

STH should check if Docker deamon is running and what is its version.

PR #52, opened 2021-08-18
Fix: Make STH fetch images ahead of starting host.

a-tylenda:
task: https://app.asana.com/0/1200747048063058/1200787625956055

Activate two BDD scenarios, which include test HUB parameters for setting runner and pre-runner docker images.


alicja-gruzdz (approved):

MichalCz:
I made a refactor of `LifeCycleAdapters` to make them available to load on host initialization so that we can replace them in hub on demand (as a first step to what @pietrzakacper suggested on making runner dev not such a hassle - and to be frank much more).

This essentially solves #32 also, and #39 if the right images are given.

To be clear: the host will not expose a port nor will it connect to manager before fetching runner and prerunner images.

MichalCz:
So I got to now have a rather peculiar error, [STH now says here](https://github.com/scramjetorg/transform-hub/runs/3389544197?check_suite_focus=true#step:11:236) that 

```
2021-08-21T14:36:57.385Z log (object:SocketServer) Supervisor connected! ID:  56d495a9-2e48-49b5-b977-000d9a937c4c <and some at signs here>
```

Dunno where do those `@` signs come from, but there's some kind of garbage. Any ideas?

patuwwy:
> Dunno where do those `@` signs come from, but there's some kind of garbage. Any ideas?

guessing... 
supervisor sends it's id as first data then muxed channels are added
host reads id and some bpmux's control data

MichalCz:
> guessing...
> supervisor sends it's id as first data then muxed channels are added
> host reads id and some bpmux's control data

Yup. Looks very much so, which is fairly bad then @patuwwy...

Is this problem also in main?

MichalCz:
Solved in another PR.

PR #53, opened 2021-08-19
Parametrise Content-Type in Runner's input stream

pietrzakacper:
Changes include:

* Sending `Content-Type` header in Host to Runner's input stream
* Reading headers in Runner's input stream
* Mapping instance `inputDataStream` according to `Content-Type` value
* Enabling logger in Runner a bit sooner to catch logs from initialization
* Disabling `checkAccepts` to allow any Content-Type value in /input endpoint

https://app.asana.com/0/1200747048063058/1200808622558080

### DIFF
--- a/packages/api-server/src/handlers/stream.ts
+++ b/packages/api-server/src/handlers/stream.ts
@@ -80,12 +80,14 @@ export function createStreamHandlers(router: SequentialCeroRouter) {
     const downstream = (
         path: string | RegExp,
         stream: StreamOutput,
-        { json = false, text = false, end: _end = false, encoding = "utf-8" }: StreamConfig = {}
+        { json: _json = false, text: _text = false, end: _end = false, encoding = "utf-8" }: StreamConfig = {}
     ): void => {
         router.post(path, async (req, res, next) => {
             try {
 
-                checkAccepts(req.headers["content-type"], text, json);
+                // ASSUMPTION: we're not using content-type anywhere else besides validation
+                // @TODO fix it
+                // checkAccepts(req.headers["content-type"], text, json);
                 if (req.headers.expect === "100-continue") res.writeContinue();
 
                 const end = checkEndHeader(req, _end);
--- a/packages/api-server/src/handlers/stream.ts
+++ b/packages/api-server/src/handlers/stream.ts
@@ -80,14 +80,15 @@ export function createStreamHandlers(router: SequentialCeroRouter) {
     const downstream = (
         path: string | RegExp,
         stream: StreamOutput,
-        { json: _json = false, text: _text = false, end: _end = false, encoding = "utf-8" }: StreamConfig = {}
+        { json = false, text = false, end: _end = false, encoding = "utf-8", checkContentType = true }: StreamConfig = {}
     ): void => {
         router.post(path, async (req, res, next) => {
             try {
 
-                // ASSUMPTION: we're not using content-type anywhere else besides validation
-                // @TODO fix it
-                // checkAccepts(req.headers["content-type"], text, json);
+                if (checkContentType) {
+                    checkAccepts(req.headers["content-type"], text, json);
+                }
+
                 if (req.headers.expect === "100-continue") res.writeContinue();
 
                 const end = checkEndHeader(req, _end);
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -44,9 +44,14 @@ export function loopStream<T extends unknown>(
         };
 
         stream.on("error", rej);
+
+        // run it in case readable was already triggered
+        onReadable();
         stream.on("readable", onReadable);
     });
 }
+
+const HEADERS_ENDING_SEQ = "\r\n\r\n";
 
 /**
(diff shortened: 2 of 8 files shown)


packages/api-server/src/handlers/stream.ts
```diff
@@ -80,12 +80,14 @@ export function createStreamHandlers(router: SequentialCeroRouter) {
     const downstream = (
         path: string | RegExp,
         stream: StreamOutput,
-        { json = false, text = false, end: _end = false, encoding = "utf-8" }: StreamConfig = {}
+        { json: _json = false, text: _text = false, end: _end = false, encoding = "utf-8" }: StreamConfig = {}
     ): void => {
         router.post(path, async (req, res, next) => {
             try {
 
-                checkAccepts(req.headers["content-type"], text, json);
```
MichalCz, line 88: We are doing checks on this, and this is a separate module that can be used to stream stuff directly. I'd add a switch that allows disabling this.

packages/runner/src/runner.ts
```diff
@@ -173,22 +173,40 @@ export class Runner<X extends AppConfig> implements IComponent {
...
-            })
-        ;
-        // eslint-disable-next-line @typescript-eslint/no-floating-promises
-        // this.inputDataStream.run();
+        this.logger.log("Input stream HELLo");
+
+        const inputStream = createReadStream(this.inputFifoPath)!;
+        const headers = await this.readInputStreamHeaders(inputStream);
+
+        this.logger.log(headers);
```
MichalCz, line 209: I understand there's more work to be done here.

pietrzakacper: yes, this is only a draft

packages/runner/src/runner.ts
```diff
@@ -16,6 +16,68 @@ type MaybeArray<T> = T | T[];
...
+                    stream.unshift(result.unconsumedData);
+                }
+
+                res(result.data);
+                break;
+            }
+        };
+
+        stream.on("error", rej);
+        stream.on("readable", onReadable);
```
MichalCz, line 47: You should execute the `onReadable` here once during loopStream call since the stream may already be readable. It's safe, as `read` will not yield any results otherwise and the body of the `while` will not be executed.

packages/runner/src/runner.ts
```diff
@@ -16,6 +16,68 @@ type MaybeArray<T> = T | T[];
...
+    });
+}
+
+/**
+ *
+ * @param stream
+ * @returns object with header key/values (header names are lower case)
+ */
+function readInputStreamHeaders(stream: Readable): Promise<Record<string, string>> {
+    const HEADERS_ENDING_SEQ = "\r\n\r\n";
```
MichalCz, line 57: This can be defined in module scope.

packages/runner/src/runner.ts
```diff
@@ -173,24 +235,36 @@ export class Runner<X extends AppConfig> implements IComponent {
...
+
+            if (contentType.endsWith("x-ndjson")) {
+                this.inputDataStream = StringStream
+                    .from(this.inputStream, { encoding: "utf-8" })
+                    .lines()
+                    .JSONParse();
+            } else if (contentType === "text/plain") {
+                this.inputDataStream = StringStream.from(this.inputStream, { encoding: "utf-8" });
+            } else if (contentType === "application/octet-stream") {
+                this.inputDataStream = DataStream.from(this.inputStream);
```
MichalCz, line 259:
```suggestion
                this.inputDataStream = BufferStream.from(this.inputStream);
```

packages/runner/src/runner.ts
```diff
@@ -203,8 +277,10 @@ export class Runner<X extends AppConfig> implements IComponent {
     }
 
     async hookupFifoStreams() {
+        // @TODO bring it back to normal
```
MichalCz, line 280: Nah, that can actually not be such a bad idea.

MichalCz: Leave it as you made it, logger will be initialized ahead of other streams so if anything goes wrong there "a chance" log will show us something.

MichalCz (commented):
Looks good, see comments here and there.

The checkAccepts should be still taken care of, otherwise some small fixes and we should be good to go. :)

packages/api-server/src/handlers/stream.ts
pietrzakacper: fixed

packages/runner/src/runner.ts
pietrzakacper: fixed

packages/runner/src/runner.ts
pietrzakacper: fixed

packages/runner/src/runner.ts
pietrzakacper: fixed

packages/runner/src/runner.ts
pietrzakacper: fixed

MichalCz (approved):

PR #69, opened 2021-08-31
fix deps

patuwwy:


alicja-gruzdz (approved):
Wow, finally a successful build :muscle:

patuwwy:
> Wow, finally a successful build 💪

not the first one :)

PR #71, opened 2021-09-01
Document environment variables.

pietrzakacper:
https://app.asana.com/0/1200834321125376/1200641162753586

### DIFF
--- a/ENV_VARS.md
+++ b/ENV_VARS.md
@@ -1,0 +1,89 @@
+### General
+---
+`PRODUCTION: boolean`
+
+---
+`DEVELOPMENT: boolean`
+
+`SCRAMJET_DEVELOPMENT: boolean`
+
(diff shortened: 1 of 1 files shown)


ENV_VARS.md
```diff
@@ -0,0 +1,89 @@
...
+### Scripts
+
+#### packsequence.js
+`OUT_DIR: string (default: dist)`
+Used to find built sourcecode.
+
+#### publish-order-dist-packages.js
+`SCRAMJET_PUBLISH_REPO: string`
+
+#### publish.js
```
MichalCz, line 76: This needs to be cleaned up.

@alicja-gruzdz please consider creating a module for env vars so that they're fetched from a single place and we can manage them.

MichalCz (approved):

PR #75, opened 2021-09-01
STH SDAdapter

patuwwy:


packages/model/src/utils/id-provider.ts
```diff
@@ -1,12 +1,11 @@
...
 export class IDProvider {
     static generate() {
-        return uuidv4();
+        return uuidv4.v4();
     }
 
-    public static isValid(id: string) {
-        return validate(id);
+    public static isValid(_id: string) {
+        return true;//uuidv4.validate(id);
```
patuwwy, line 9: to be fixed

packages/runner/src/runner.ts
```diff
@@ -174,24 +187,21 @@ export class Runner<X extends AppConfig> implements IComponent {
...
+        this.logger.log(`Content-Type: ${contentType}`);
+
+        mapToInputDataStream(this.inputStream!, contentType).pipe(this.inputDataStream!);
     }
 
     async hookupOutputStream() {
         this.outputStream = createWriteStream(this.outputFifoPath);
         this.outputDataStream
-            .JSONStringify()
+            //JSONStringify()
```
patuwwy, line 204: TBD

packages/runner/src/input-stream.ts
```diff
@@ -73,7 +73,10 @@ export function mapToInputDataStream(stream: Readable, contentType: string): Dat
     if (contentType.endsWith("x-ndjson")) {
         return StringStream
             .from(stream, { encoding: "utf-8" })
-            .JSONParse(true);
+            .JSONParse(true)
+            .catch((error: any) => {
+                console.error("mapToInputDataStream", error);
+            });
```
MichalCz, line 79:
```suggestion
                // TODO: we should be doing some error handling here: remove the stream, mark as bad, kill the instance maybe?
            });
```

packages/runner/src/runner.ts
MichalCz: This should be defined by:

1) Package.json
2) Return value of the last item in sequence

Where 2 overrides 1.

packages/runner/src/runner.ts
```diff
@@ -542,6 +589,14 @@ export class Runner<X extends AppConfig> implements IComponent {
                         ? this.outputStream
                         : this.outputDataStream
                 );
+
+            MessageUtils.writeMessageOnStream(
+                [RunnerMessageCode.PANG, {
+                    provides: intermediate.topic || "",
+                    contentType: intermediate.contentType || ""
```
MichalCz, line 596:
```suggestion
                    provides: intermediate.topic ? `${intermediate.topic}` : undefined,
                    isRaw: !!intermediate.isRaw,
                    contentType: intermediate.contentType ? `${intermediate.contentType}` : undefined
```

MichalCz (approved):

packages/runner/src/runner.ts
alicja-gruzdz: thanks, we certainly need to handle it but we need a consultation with you about it. It will also cause changes that need to be well tested. Therefore, we propose to solve it in a separate PR:
https://app.asana.com/0/1199599848882722/1200949562077777

a-tylenda (approved):

PR #81, opened 2021-09-14
Send and get topic data with CLI

patuwwy:
send contents of data.json file to topic named "abc"
```si topic send abc data.json```

stdin will be used when filename is not provided 

get data from topic named "abc"
```si topic get abc```


a-tylenda (approved):

a-tylenda:
Docs must be updated with those two new commands.

PR #92, opened 2021-09-25
Colorize logger function name, update logs.

patuwwy:
image

### DIFF
--- a/packages/adapters/src/instance-adapter.ts
+++ b/packages/adapters/src/instance-adapter.ts
@@ -275,6 +275,8 @@ IComponent {
 
             this.logger.log("Container exited.");
 
+            await defer(10000);
+
             if (statusCode > 0) {
                 throw new SupervisorError("RUNNER_NON_ZERO_EXITCODE", { statusCode });
             } else {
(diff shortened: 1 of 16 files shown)


packages/adapters/src/instance-adapter.ts
```diff
@@ -270,14 +268,12 @@ IComponent {
...
 
-        this.logger.debug(`Container is running (${containerId})`);
+        this.logger.log(`Container is running (${containerId}).`);
 
         try {
             const { statusCode } = await this.dockerHelper.wait(containerId);
 
-            this.logger.debug("Container exited");
-
-            await defer(10000);
```
patuwwy, line 280: This should be removed in separate PR

alicja-gruzdz (approved):
:paintbrush: :art:

a-tylenda (approved):

packages/adapters/src/instance-adapter.ts
alicja-gruzdz:
```suggestion
            await defer(10000);
```

PR #95, opened 2021-09-30
Update FreePortsFinder to search ports in range

patuwwy:


MichalCz (approved):

patuwwy:
> This method dosn't detect docker container exposed ports.... Node can set up HTTP server on port 30000 even if there is running container exposing on the same port

resolved

PR #98, opened 2021-10-02
Fix/end stream cont

patuwwy:
Fixes for 
- hanging connections, 
- sending stream with end
- lost first output

and many more

### DIFF
--- a/packages/host/src/lib/host.ts
+++ b/packages/host/src/lib/host.ts
@@ -11,7 +11,7 @@ import { SocketServer } from "./socket-server";
 
 import { unlink, access as access } from "fs/promises";
 import { IncomingMessage, ServerResponse } from "http";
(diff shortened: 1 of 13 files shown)


alicja-gruzdz (approved):

packages/host/src/lib/host.ts
```diff
@@ -357,31 +355,32 @@ export class Host implements IComponent {
...
                 );
             }
 
             if (data.provides) {
                 notifyCPM = true;
 
                 this.logger.log("Sequence provides data: ", data);
-                this.serviceDiscovery.addData(
-                    csic.getOutputStream()!,
+                const topic = this.serviceDiscovery.addData(
```
MichalCz, line 377: Nice. :)

MichalCz (approved):

PR #134, opened 2021-11-08
Change CLI config filename

pietrzakacper:
Changes include: 

* Renaming CLI config filename so it's not shared with CPM CLI
* Minor typing improvements

### DIFF
--- a/packages/cli/src/lib/config.ts
+++ b/packages/cli/src/lib/config.ts
@@ -10,7 +10,7 @@ const defaultConfig = {
 };
 
 type Config = typeof defaultConfig;
-const location = resolve(homedir(), ".scramjetrc");
+const location = resolve(homedir(), ".scramjet_sth_cli_rc");
 
 let currentConfig: Config;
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/config.ts
```diff
@@ -10,7 +10,7 @@ const defaultConfig = {
 };
 
 type Config = typeof defaultConfig;
-const location = resolve(homedir(), ".scramjetrc");
+const location = resolve(homedir(), ".scramjet_sth_cli_rc");
```
patuwwy, line 13: Could we use `.sth-cli.json` similar to `/tmp/sth-id.json` if this is json file?

patuwwy (changes requested):

alicja-gruzdz (approved):

packages/cli/src/lib/config.ts
pietrzakacper: fixed

patuwwy (approved):

a-tylenda (approved):

PR #135, opened 2021-11-09
Type core STH Rest API endpoints

pietrzakacper:
Changes include:
* Introducing `STHRestApi` namespace with endpoints responses and creating structure for future RestAPI tests.
* Aligning zIInstancez and `ISequence` with typing guidelines (`interface ISequence` -> `type Sequence`)
* Removing `Sequence` class that acted as object mapping and contained typing errors (`instances: any[] = []`)
* Other small refactors and fixes

#### Decisions to be made
I've implemented some basic type-layer error mechanism for `StartSequence` and `DeleteSequence`
using a mechanism based on what was already in place: `ReasonPhrases` from 'http-status-codes' package.
It's an easy way to check for error response on the frontend:

```typescript
export type OpResponse< PayloadType extends Record<string, unknown>> =
| (PayloadType & { opStatus: ReasonPhrases.OK })
| { opStatus: Exclude<ReasonPhrases, ReasonPhrases.OK>, error?: unknown }

declare const response: OpResponse<{id : string}>

response.id // error, we need to check for opStatus first

if(response.opStatus === ReasonPhrases.OK) {
  response.id // id is now available on the type
} else {
  reponse.error // here we have some error that we log
}
```
The downside of this approach is not always HTTP statuses will be enough to express business logic error in our platform.
In the future we might want to maintain our own set of error codes for client-side error handling.
However I am not sure how much emphasis will be put on good error handling in the panel at this stage of development so I just wanted to set a direction :wink:


patuwwy (approved):
LG

a-tylenda (approved):

packages/types/src/messages/describe-sequence.ts
```diff
@@ -27,7 +27,7 @@ export type FunctionDefinition = {
      */
     name?: string;
     /**
-     * Addtional description of the function
+     * Additional description of the function
```
a-tylenda, line 30: There are two more typos in this file, maybe you could fix them while you are here 😉 
- L6:  funcion -> function
- L8: neighbours -> neighbors 

thanks 💪🏼

PR #151, opened 2021-12-01
Enable starting runner with ts-node in --no-docker mode

pietrzakacper:
Other than that I've refactored some things that didn't get into the original PR.

One change worth noting is choice of JSON decoding library.
[ts.data.json](https://github.com/joanllenas/ts.data.json)

I've had it battle tested in my previous work and it works like a charm. I've had a chance to play with similar libraries and this one is imho the most intuitive and robust solution. 

### Verify
To verify we can just run any BDD test with `SCRAMJET_SPAWN_TS=1 NO_DOCKER=1` e.g.
```
SCRAMJET_SPAWN_TS=true NO_DOCKER=true SCRAMJET_TEST_LOG=true DEVELOPMENT=true yarn test:bdd --name="PT-004 TC-001 Checksum of JSON payload"
```
and verify in logs that both `supervisor` and `runner` were run using TS source code
```
2021-12-01T18:18:59.941Z info (object:CSIController:40ea8abe-4e6c-4835-857b-33aa33f8d0cd) Spawning supervisor with command: [
  '/home/kpietrzak/Projects/scramjet-cpm-dev/sth/packages/supervisor/src/bin/supervisor.ts',
  '40ea8abe-4e6c-4835-857b-33aa33f8d0cd',
  '/tmp/scramjet-socket-server-path-34947'
]

(...)

2021-12-01T18:19:02.805Z log (object:ProcessInstanceAdapter) Spawning Runner process with command [
  'ts-node',
  '/home/kpietrzak/Projects/scramjet-cpm-dev/sth/packages/runner/src/bin/start-runner.ts'
] and envs:  {
  FIFOS_DIR: '/tmp/fifosAukCK0',
  SEQUENCE_PATH: '/home/kpietrzak/.scramjet_sequences/535eeaa8-6349-4f64-a0d8-9992a14e6509/index.js'
}
```

### DIFF
--- a/packages/host/src/lib/csi-controller.ts
+++ b/packages/host/src/lib/csi-controller.ts
@@ -137,6 +137,8 @@ export class CSIController extends EventEmitter {
         this.superVisorProcess = spawn(executable, command, {
             env: {
                 PATH: process.env.PATH,
+                DEVELOPMENT: process.env.DEVELOPMENT,
+                PRODUCTION: process.env.PRODUCTION,
                 NO_DOCKER: this.config.noDocker.toString()
             }
         });
--- a/packages/host/src/lib/csi-controller.ts
+++ b/packages/host/src/lib/csi-controller.ts
@@ -122,7 +122,6 @@ export class CSIController extends EventEmitter {
     }
 
     startSupervisor() {
-        // eslint-disable-next-line no-extra-parens
         const isTSNode = !!(process as any)[Symbol.for("ts-node.register.instance")];
         const supervisorPath = require.resolve("@scramjet/supervisor");
 
@@ -137,6 +136,7 @@ export class CSIController extends EventEmitter {
 
         this.superVisorProcess = spawn(executable, command, {
             env: {
+                ...process.env,
                 NO_DOCKER: this.config.noDocker.toString()
             }
         });
--- a/packages/host/src/lib/csi-controller.ts
+++ b/packages/host/src/lib/csi-controller.ts
@@ -136,7 +136,7 @@ export class CSIController extends EventEmitter {
 
         this.superVisorProcess = spawn(executable, command, {
             env: {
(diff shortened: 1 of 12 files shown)


packages/host/src/lib/csi-controller.ts
```diff
@@ -137,6 +136,7 @@ export class CSIController extends EventEmitter {
 
         this.superVisorProcess = spawn(executable, command, {
             env: {
+                ...process.env,
```
MichalCz, line 139: @pietrzakacper I'm not sure if I wouldn't just remove this and pass ONLY the env we want in the runner. What are we actually using besides no-docker? I guess this will anyway be removed when we get rid of Supervisor?

MichalCz (approved):

packages/host/src/lib/csi-controller.ts
pietrzakacper: The reason it was needed is that PATH env wasn't set, so ts-node binary couldn't be found. 
I changed it to explicitly pass only PATH :wink:

pietrzakacper: I also added DEVELOPMENT & PRODUCTION envs since it's quite useful to pass them

alicja-gruzdz: @pietrzakacper yep, good idea to have DEV & PROD envs there, we use them a lot for :bug:

alicja-gruzdz:
Thanks for adding the verify step!

alicja-gruzdz (approved):
Thanks, it will be very useful to our rapid debugging. In fact, I feel this new mode and I will become best friends ;)

Supervisor:
Zrzut ekranu z 2021-12-02 15-37-46


Runner
Zrzut ekranu z 2021-12-02 15-38-13
er:

PR #157, opened 2021-12-08
TCP Runner

pietrzakacper:
## Overview
This PR aims to solve these challenges:
1. We are preparing for deploying STH to Kubernetes so we cannot use local filesystem (FIFOs, process STDIO) to handle `host` <-> `runner` communication. 
2. We want to be supporting sequences written in multiple languages which requires support for multi language `runner` implementations. So we cannot use NodeJS specific libraries for communication like `bpmux`.
3. Our architecture, due to historical reasons, contains one extra process for every instance - `supervisor`. Which is introducing unnecessary complexity.

### Old architecture (`host` <--bpmux--> `supervisor` <--fifos--> `runner`)
image
### New architecture  (`host` <--tcp--> `runner`)
image
#### Removing supervisor process
***Before*** this PR, for each instance `host` would spawn a `supervisor` process which would spawn `runner` process (using appropriate `instanceAdapter`). The role of supervisor was to translate TCP communication with `host` to FIFO communication with `runner`. And also to handle the lifecycle of `runner` process such as killing it if it doesn't exit after sequence completion.
It turned out to contribute to unnecessary complexity of the platform so a natural move was to get rid of it.
***Currently*** `host` directly spawns and communicates with `runner` process over TCP. 

#### Replacing single BPMux connection with raw TCP connection per channel
***Before*** `host` would receive single TCP connection per instance that would service all channels (control, input, output etc.). To prevent the channels from blocking each other when one of them experiences significant load we used `bpmux` library.
However this solution wasn't very portable to other languages. It would require us to write a `bpmux` implementation for every technology we want the `runner` to be implemented in.
***Currently*** We decided to use raw TCP connections to handle `host` <-> `runner` communication. Intitially the idea was to leverage the duplex nature of TCP connections and group the channels by the opposite data transfer directions e.g monitor/control, input/output. But we (not so) quickly realized that whenever a channel, be it input channel closes an underlying connection the output channel would be closed to. To prevent added complexity from handling such scenarios we decided to simply have 8 TCP connections, one for each channel (stdio, stdout, stderr, input, output, control, monitor, log).

#### and some smaller changes like ...:
* Getting rid of `snapshotOnError` concept (It wasn't even implemented and it makes almost no sense without FIFOs)
* Getting rid of `ForceKeepAlive` message. (it wasn't used anywhere)
* other small cleanups and refactors

### TODO
* [ ] Replace `docker_network_mode = 'host'` with port forwarding possibly
* [ ] Pass IP of instances server apart from Runner (might leave that for Kuberenetes adapter)
* [ ] Unit tests for Runner
* [x] Refactor CSIController (it got merged with Lifecycle controller and probably the logic is much more complex than it needs to be)
* [x] Send stdio over TCP as well
* [x] Investigate bpmux erros Write after end and maybe suppress them (possibly the better choice is just to create an issue as this doesnt seem to be serious and it was already happening before this PR)
* [ ] Discuss how Runner quitting delay should work (in every cases: output stream end, error, stop, kill etc.)
* [x] Investigate `makeSnapshotOnError` which is always false (possibly just create an issue)
* [x] Investigate `'(node:2557) UnhandledPromiseRejectionWarning: Error: carrier stream ended before end message received\n` in `E2E-006 TC-001`
* [x] Support changes from here: https://github.com/scramjetorg/transform-hub/commit/81a86d76c40da7ba5613097fe4ae051ba9df41e4 
* [ ] Runner Context cleanup 
* [ ] Support replacement communication protocols in InstanceAdapter
* [ ] Rethink number of connections
* [ ] Handle connection interruption
* [ ] Update docs
* [ ] Cleanup connected instances in socket-server


packages/runner/src/bin/start-runner-local.ts
```diff
@@ -1,109 +1,109 @@
...
+// #!/usr/bin/env node
+// /* eslint-disable dot-notation */
+
+// import { RunnerMessageCode } from "@scramjet/symbols";
+// import { AppConfig, HandshakeAcknowledgeMessage, ReadableStream } from "@scramjet/types";
+
+// import { Runner } from "../runner";
+// import { stdout } from "process";
+// import * as fs from "fs";
+// import * as path from "path";
```
pietrzakacper, line 10: want to write those after the review since Runner is prone to change rn :wink:

packages/runner/package.json
```diff
@@ -8,7 +8,7 @@
     "build:docker": "docker build -t scramjetorg/runner:$npm_package_version -f Dockerfile ../../",
     "build:docs": "typedoc",
     "start": "ts-node ./src/index",
-    "test": "nyc ava",
+    "test": "echo \"Skip for now\" #nyc ava",
```
pietrzakacper, line 11: Yeah, later these should run

alicja-gruzdz:
Fantastic. With this brand new Runner, we jetted to the cloud space :rocket:  next stop Kubernetes!
:clap:

alicja-gruzdz:
Good that you removed the `snapshotOnError`. We have not come to implementing this feature as there were other pressing issues. This feature was intended to make a snapshot of a container in case of an Instance failure and make it available to a user for inspection.
As far as I understand now we tackle this differently. We will support saving and restoring the Instance state with Save&Restore mechanism so I'm not sure we're planning on creating snapshots of containers. I know the functionalities are not exactly the same, so I'd prefer if @MichalCz and @luccam-scr  confirmed this.

alicja-gruzdz:
Like `snapshotOnError`, `ForceKeepAlive` has never been implemented. The [idea](https://github.com/scramjetorg/transform-hub/blob/68f9e119bbc77975bd297214cd88869dbc6cfe89/packages/types/src/messages/confirm-health.ts#L4) behind it was that since we terminate unresponsive Instances (the ones that do not emit the health message), if an Instance stops to emit the health messages we want to give it a last chance to save its life by sending to it the `FORCE_CONFIRM_ALIVE` message (probably not the clearest name for it). If Instance replies to `FORCE_CONFIRM_ALIVE` we give it more lifespan and wait for incoming health messages, if not, we cold-bloodedly kill it.

packages/adapters/src/process-instance-adapter.ts
```diff
@@ -31,101 +23,19 @@ class ProcessInstanceAdapter implements
...
-        this.controlStream = new DelayedStream();
-        this.loggerStream = new DelayedStream();
-        this.inputStream = new DelayedStream();
-        this.outputStream = new DelayedStream();
-
         this.logger = getLogger(this);
     }
 
     async init(): Promise<void> {
         // noop
```
alicja-gruzdz, line 35: I see that in the Instance Adapters we always have the init function empty. I wonder whether we'll some initiation to do in the Kubernetes Instance Adapter. If we come to it and see we don't have anything to initiate, maybe we could remove it.

packages/host/src/lib/socket-server.ts
```diff
@@ -1,116 +1,81 @@
...
-                this.emit("connect", {
-                    id,
-                    streams
+                connection.on("error", (err) => {
+                    this.logger.error(`Error on instance ${id} in stream ${channel}`, err);
                 });
+
+                this.logger.info(`Connection on channel: ${channel}`);
+
+                // @TODO check it it runner[channel] was null before if not throw
```
alicja-gruzdz, line 65: That is a rather good idea :)

packages/runner/src/csh-client.ts
```diff
@@ -0,0 +1,118 @@
...
+
+                        return connection;
+                    });
+                })
+        );
+
+        this._streams = openConnections as HostOpenConnections;
+    }
+
+    async disconnect() {
```
alicja-gruzdz, line 60: Perhaps it would be useful to add some logging here, we often debug the end of stream events.

alicja-gruzdz (approved):

alicja-gruzdz:
A question: when we send a kill message to an Instance, a Runner should exit, but do we have any backup mechanism to kill the Instance if the Runner for any reason fails to exit? (for both Instance Adapters). @MichalCz asked this question today on demo event, I wasn't sure if we have it currently implemented. It's not directly related to this PR, but raising this issue while we work on Host<->Runner communication changes.

packages/adapters/src/process-instance-adapter.ts
pietrzakacper: I'll keep that in mind

packages/runner/src/csh-client.ts
pietrzakacper: added

packages/host/src/lib/socket-server.ts
pietrzakacper: fixed

patuwwy (approved):

packages/host/src/lib/socket-server.ts
alicja-gruzdz: cool, thanks :)

bdd/features/e2e/E2E-003-kill.feature
```diff
@@ -7,7 +7,6 @@ Feature: Kill e2e tests
         And instance started
         And wait for instance healthy is "true"
         And get runner PID
-        And send kill message to instance
```
MichalCz, line 10: Why is this removed? How are we testing if the sequence responds to kill?

MichalCz (changes requested):

bdd/features/e2e/E2E-003-kill.feature
pietrzakacper: need to look into that and see why i removed it xd good catch :wink:

pietrzakacper: ok, brought it back, It works. must have been accidental when I was improving some tests to allow the sequence to exit on it's own, without killing it.

MichalCz (approved):

PR #166, opened 2021-12-14
Add missing export in @scramjet/types

a-tylenda:
This change will need to be released


alicja-gruzdz (approved):
:clap: for finding it!

Issue #165, opened 2021-12-14
CLI command: "si instance input <instance-id>" doesn't read data from stdin


a-tylenda:
**Describe the bug**
CLI command `si instance input -h` says:
```
Usage: si instance input [options] <id> [<file>]

send file to input, if file not given the data will be read from stdin

Options:
  -t, --content-type <value>  Content-Type (default: "text/plain")
  -h, --help                  display help for command
```
When I tried to send some data to my running sequence with this command: `$ si instance input 7a1ffd59-9d1a-4e8f-a246-020124803931` it returned an error:
```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:371:5)
    at validateString (node:internal/validators:119:11)
    at resolve (node:path:1098:7)
    at getReadStreamFromFile (/home/atylenda/.nvm/versions/node/v16.6.1/lib/node_modules/@scramjet/cli/lib/common.js:84:49)
    at Command.<anonymous> (/home/atylenda/.nvm/versions/node/v16.6.1/lib/node_modules/@scramjet/cli/lib/commands/instance.js:63:119)
    at Command.listener [as _actionHandler] (/home/atylenda/.nvm/versions/node/v16.6.1/lib/node_modules/@scramjet/cli/node_modules/commander/lib/command.js:488:17)
    at /home/atylenda/.nvm/versions/node/v16.6.1/lib/node_modules/@scramjet/cli/node_modules/commander/lib/command.js:1227:65
    at Command._chainOrCall (/home/atylenda/.nvm/versions/node/v16.6.1/lib/node_modules/@scramjet/cli/node_modules/commander/lib/command.js:1144:12)
    at Command._parseCommand (/home/atylenda/.nvm/versions/node/v16.6.1/lib/node_modules/@scramjet/cli/node_modules/commander/lib/command.js:1227:27)
    at Command._dispatchSubcommand (/home/atylenda/.nvm/versions/node/v16.6.1/lib/node_modules/@scramjet/cli/node_modules/commander/lib/command.js:1050:25)
```
I also tried this way: `si instance input 7a1ffd59-9d1a-4e8f-a246-020124803931 John`, also didn't work out and returned an error:
```
Error: File "John" not found.
    at /home/atylenda/.nvm/versions/node/v16.6.1/lib/node_modules/@scramjet/cli/lib/common.js:86:15
[...]

PR #177, opened 2021-12-21
Do not check if file exists when filename has not been provided

patuwwy:
in `si inst input` command

after fixing this I've found another bug when i've tried to send from stdin and then send a file contents

Fixed it in the way that subsequent requests to instance `input` don't send `content-type` to `Runner` which actually doesn't parse headers more than once.


alicja-gruzdz (approved):

patuwwy:
Fixes [#165](https://github.com/scramjetorg/transform-hub/issues/165)

a-tylenda:
veryfied 
- stdin ✅ 
- `.txt` file ✅ 
image

a-tylenda (approved):

PR #182, opened 2021-12-29
Add more scenarios to new workflow

a-tylenda:
This pull request is related to PR [#183](https://github.com/scramjetorg/transform-hub/pull/183), where a new workflow for npm packages (@scramjet/sth & @scramjet/cli) was created. In this PR the number of tested scenarios was extended and adjusted accordingly.

### DIFF
--- a/bdd/step-definitions/e2e/cli.ts
+++ b/bdd/step-definitions/e2e/cli.ts
@@ -7,14 +7,10 @@ import * as fs from "fs";
 import { getStreamsFromSpawn } from "../../lib/utils";
 import { expectedResponses } from "./expectedResponses";
 
-// const si = process.env.SCRAMJET_SPAWN_TS
-//     ? ["npx", "ts-node", "../packages/cli/src/bin/index.ts"]
-//     : ["node", "../dist/cli/bin"]
-// ;
-
+// eslint-disable-next-line no-nested-ternary
 const si = process.env.SCRAMJET_SPAWN_JS
-    ? ["node", "../dist/cli/bin"]
-    : ["si"]
+    ? ["node", "../dist/cli/bin"] : process.env.SCRAMJET_SPAWN_TS
+        ? ["npx", "ts-node", "../packages/cli/src/bin/index.ts"] : ["si"]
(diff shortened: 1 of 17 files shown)


MichalCz (approved):

bdd/step-definitions/e2e/cli.ts
```diff
@@ -7,10 +7,16 @@ import * as fs from "fs";
...
 
-const si = process.env.SCRAMJET_SPAWN_TS
-    ? ["npx", "ts-node", "../packages/cli/src/bin/index.ts"]
-    : ["node", "../dist/cli/bin"]
+// const si = process.env.SCRAMJET_SPAWN_TS
+//     ? ["npx", "ts-node", "../packages/cli/src/bin/index.ts"]
+//     : ["node", "../dist/cli/bin"]
+// ;
+
+const si = process.env.SCRAMJET_SPAWN_JS
```
alicja-gruzdz, line 15: I'd add it to our ENV Bible: https://github.com/scramjetorg/transform-hub/blob/release/0.13/ENV_VARS.md :)
so the way I understand it, our tests which are not run with `SCRAMJET_SPAWN_JS` will be testing the CLI installed from npm (the one installed locally)?
Hmm maybe we could mention it somewhere in the tests. I just feel I will find myself debugging the new fixes in CLI and not understanding why they still don't work :smile:

alicja-gruzdz (approved):

bdd/step-definitions/e2e/cli.ts
alicja-gruzdz: as an alternative perhaps allowing one of the three: ts/js/npm?

a-tylenda: new ENV added to our Bible 😉 I also added short description which says that when `SCRAMJET_SPAWN_JS` is set to `false` cli commands are executed from npm package with `si` command.

alicja-gruzdz: excellent, thanks :)

PR #191, opened 2022-01-03
Add code comments in api-server module

patuwwy:


packages/api-server/src/handlers/op.ts
```diff
@@ -105,11 +126,14 @@ export function createOperationHandler(router: SequentialCeroRouter) {
         };
 
         switch (method) {
+        case "post":
+            router.post(path, handler);
+            break;
         case "delete":
-            router.delete(path, handler);
+            router.post(path, handler);
```
patuwwy, line 133:
```suggestion
            router.delete(path, handler);
```

patuwwy: fix

packages/types/src/api-expose.ts
```diff
@@ -116,11 +123,23 @@ export interface APIBase {
...
     duplex(
         path: string | RegExp,
         callback: (stream: Duplex, headers: IncomingHttpHeaders) => void
     ): void;
 
+    /**
+     * Allows to register middlewares for specific paths, for all HTTP methods.
+     *
+     * @param path
+     * @param middlewares
```
patuwwy, line 141:
```suggestion
     * @param {Middleware[]} middlewares
```

packages/types/src/api-expose.ts
```diff
@@ -116,11 +123,23 @@ export interface APIBase {
...
+     */
     duplex(
         path: string | RegExp,
         callback: (stream: Duplex, headers: IncomingHttpHeaders) => void
     ): void;
 
+    /**
+     * Allows to register middlewares for specific paths, for all HTTP methods.
+     *
+     * @param path
```
patuwwy, line 140:
```suggestion
     * @param {string} path
```

alicja-gruzdz (approved):

packages/api-server/src/lib/data-extractors.ts
```diff
@@ -3,14 +3,29 @@ import { ServerResponse, IncomingMessage } from "http";
...
 }
 
+/**
+ * Returns result of the function call if object is a function or the object itself.
+ *
+ * @param {any} object Object.
+ * @param {IncomingMessage} req Request.
+ * @param {ServerResponse} res Response.
+ * @returns {Promise<any>} Promise resolving to the result of the function call or to the object itself.
+ */
```
f1ames, line 28: It bothers me a bit since this function is almost identical to the one above (apart from additional param and return type) - and since it still accepts `any` as first param the possible `results` would be the same as in `getObject()`? And `getWritable` name seems to be somehow inconsistent with docs / what this function does.

I know that it is probably used in a different context, but without it, it's quite confusing. And I know that this PR is about docs, not code :see_no_evil:

packages/api-server/src/lib/data-extractors.ts
```diff
@@ -19,6 +34,14 @@ export async function getWritable(object: any, req: IncomingMessage, res: Server
     return object;
 }
 
+/**
+ * Identifies type of given object. If object is a function, it will be called.
+ *
+ * @param req Request object.
+ * @param res Response object.
+ * @param stream Object to identify.
+ * @returns Readable if object is a stream or results of function call if object is a function.
```
f1ames, line 43: Looking at the description and the code I'm not sure :thinking: If `stream` is a `Promise` or a `function` it gets resolved/called but its result is passed to `getStream` again - so I guess the purpose here is to get `stream` instance at the end or throw error? WDYT?

f1ames (changes requested):
There is never too much docs :smile: :+1: 

I have to minor suggestions for two functions as they seem quite confusing for me :thinking:

packages/api-server/src/lib/data-extractors.ts
```diff
@@ -19,6 +34,14 @@ export async function getWritable(object: any, req: IncomingMessage, res: Server
     return object;
 }
 
+/**
+ * Identifies type of given object. If object is a function, it will be called.
```
patuwwy, line 38:
```suggestion
 * Returns readable stream from given object. If object is a function it will be called and it should return a stream.
```

packages/api-server/src/lib/data-extractors.ts
```diff
@@ -3,14 +3,29 @@ import { ServerResponse, IncomingMessage } from "http";
...
 export async function getObject(object: any, req: IncomingMessage): Promise<any> {
     if (typeof object === "function") {
-        return await object(req);
+        return object(req);
     }
 
     return object;
 }
 
+/**
```
patuwwy, line 21:
```suggestion
// @TODO: Refactor.
/**
```

f1ames (approved):
WFM now :+1:

PR #194, opened 2022-01-05
Improve TCP Runner

pietrzakacper:
This PR aims to iterate on some issues introduced in #157 

Changes include:

* Cleaning up Runner class from omnipresent `if(this.context === undefined)`
* Use typed event emitter (using [typed-emitter](https://www.npmjs.com/package/typed-emitter) with zero runtime code)
* Using `bridge` network mode instead of `host` for `DockerInstanceAdapter`
* Decided to leave 8 connections for now. We could be bundling opposite data streams (input/output, monitor/control etc.) in duplex connection but we would only decrease the number of connections to 5 and the code would be a little more complex.  I created an issue though https://github.com/scramjetorg/transform-hub/issues/201
* Handle any connection errors by crashing Runner. This simplifies debugging, though later we would probably like to have some recovery mechanism in place.
* Remove Runner unit tests. They were outdated after significant changes casued by moving to TCP adn eliminating supervisor. They were written when we didn't have such coverage of BDD tests. I guess it's a good part of the app to be unit tested, especially from the host communication and user's sequence perspective. I think a better moment to have a set of thorough unit tests would be after our architecture stops undergoing rapid changes. and yes I'm also lazy.


MichalCz (approved):

patuwwy (approved):
LG

alicja-gruzdz (approved):

PR #205, opened 2022-01-14
Add last uploaded id alias in cli

MichalCz:
This PR adds CLI functionality that replaces `-` argument for the last item the user interacted with or `select`ed.

Adds `-` functionality on all commands - the sequence can be set by running `si seq|inst select <id>` or when uploading, packing or starting sequences.

Test:

`NO_DOCKER=1 yarn test:bdd --name "E2E-010 TC-023"`

A bit more elaborate test from command line:

```sh
yarn build
npm i -g ./dist/cli
si pack dist/reference-apps/hello-alice-out
si seq send -
si seq start -
si seq info -
```

Have fun.

### DIFF
--- a/bdd/features/e2e/E2E-010-cli.feature
+++ b/bdd/features/e2e/E2E-010-cli.feature
@@ -198,3 +198,25 @@ Feature: CLI tests
         And I get the second instance output
         Then confirm data named "hello-input-out-10" received
         * stop host
+
+    @ci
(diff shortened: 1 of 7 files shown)


iaforek (approved):

a-tylenda:
nice nice noooice 🥇 

verified:

image

image

a-tylenda (approved):

PR #208, opened 2022-01-17
Reenable STH to be run inside a docker container

### LINKED ISSUE #207: Docker instance adapter incorrectly assumes "docker0" interface is always available
## Describe the bug

Docker instance adapter assumes that `docker0` network interface is always available thus using it to communicate with docker instance.

https://github.com/scramjetorg/transform-hub/blob/8ba33eb25df6a6deb8a6c8245182c2411c686e66/packages/adapters/src/docker-instance-adapter.ts#L122-L138

There is a fallback hardcoded `IP` address but it doesn't really solve the issue of `docker0` being absent.

We have stumbled upon this issue while running SCP test on CI where `docker0` was not available and fallback address didn't made too sense there too.

## Steps to Reproduce

1. Run all SCP tests with `sth` bumped at least to `0427e84` commit (like in this PR https://github.com/scramjetorg/scramjet-cpm-dev/pull/135).

**Expected behavior**

All tests green.

**Actual behavior**

Some tests timeouts since docker instance adapter cannot connect to docker instances (or tho other way around :thinking:).

pietrzakacper:
After migrating Runner to TCP from FIFOs we are now more dependent on the network configuration of STH.
To take control of that we create our own Docker bridge network (`transformhub0`) during Host initialization.

Manager works properly too: https://github.com/scramjetorg/scramjet-cpm-dev/pull/135
This should fix #207

### DIFF
--- a/packages/adapters/src/docker-instance-adapter.ts
+++ b/packages/adapters/src/docker-instance-adapter.ts
@@ -211,7 +211,7 @@ IComponent {
 
         streams.stderr.on(
             "data",
-            data => this.logger.error("RUNNER DOCKER error", data));
+            data => this.logger.error("RUNNER DOCKER error", data.toString()));
         streams.stdout.pipe(process.stdout);
 
         this.resources.containerId = containerId;
--- a/packages/adapters/src/docker-instance-adapter.ts
+++ b/packages/adapters/src/docker-instance-adapter.ts
@@ -18,7 +18,6 @@ import { DockerodeDockerHelper } from "./dockerode-docker-helper";
 import { DockerAdapterResources, DockerAdapterRunPortsConfig, DockerAdapterVolumeConfig } from "./types";
 import { FreePortsFinder, defer } from "@scramjet/utility";
 import * as fs from "fs/promises";
-import { randomUUID } from "crypto";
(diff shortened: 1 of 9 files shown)


packages/adapters/src/docker-instance-adapter.ts
```diff
@@ -26,20 +28,22 @@ class DockerInstanceAdapter implements
 ILifeCycleAdapterMain,
 ILifeCycleAdapterRun,
 IComponent {
-    private dockerHelper: IDockerHelper;
+    private dockerHelper: DockerodeDockerHelper;
```
patuwwy, line 31: ```IDockerHelper``` was designed to easy replace ```Dockerode``` with other lib to communicate with Docker.
What is the reason for change ?

pietrzakacper: it was just for faster iterations. I will refactor it so it's encapsulated in the Interface

alicja-gruzdz (approved):
cool, I didn't see any issues except the question I asked in priv about create network (clear now, thanks). Please let me know if you want me to verify it manually or in this PR it will be done automatically in Bump PR in repo cpm.

PR #209, opened 2022-01-17
Introduce ObjLogger

patuwwy:
This PR introduces Object Logger (ObjLogger) and replaces old plain text Logger. 

The main assumption of the object logger is to send logs in the form of JSON objects. ObjLogger also enables the transfer of logs from one logger to another and automatic enrichment of logs with information about the components through which the logs pass.

image

Added option to filter logs showing in terminal

```
Usage: sth [options]

Options:
  -L, --log-level <level>         Specify log level (default: "trace")
```


packages/host/src/lib/host.ts
```diff
@@ -72,6 +76,11 @@ export class Host implements IComponent {
      */
     logger: Logger;
 
+    /**
+     * Instance of class providing logging utilities.
+     */
+    objLogger: ObjLogger;
```
f1ames, line 82: Would `logger` be still used for something, e.g. `/log` endpoint or we plan to get rid of it entirely?

f1ames (changes requested):
Logs look really crisp now! :star_struck: Good job :+1: 

image

As for clarifications:

For me the question is if we want to keep two separate loggers or merge them into one and then control per output stream how something is logged (json vs text). Or just get rid of `Logger` since `objLogger` already outputs text to stdout :thinking:  Maybe it's not for now :thinking: but still worth considering.
---

There are some classes where (`process-sequence-adapter.ts` and `sequence-adapter.ts`):

```
objLogger: IObjectLogger;
```

is used and others where it's like:

```
objLogger: ObjLogger;
```

Minor thing but seems inconsistent :thinking:

packages/obj-logger/src/obj-logger.ts
```diff
@@ -0,0 +1,121 @@
...
+
+                this.write(entry.level!, entry, ...entry.data || []);
+            });
+
+        this.outLogStream.pipe(this.output);
+    }
+
+    write(level: LogLevel, entry: LogEntry | string, ...optionalParams: any[]) {
+        if (ObjLogger.levels.indexOf(level) <= ObjLogger.levels.indexOf(this.logLevel)) {
+            //return;
```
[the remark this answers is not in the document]
patuwwy: yup :)

patuwwy:
@f1ames, ty for cr
previous logger is going to be totally removed / replaced with the new one.
this PR is still missing a commit removing all `this.logger.log`

packages/obj-logger/src/obj-logger.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    error(entry: LogEntry | string, ...optionalParams: any[]) {
+        this.write("ERROR", entry, ...optionalParams);
+    }
+
+    debug(entry: LogEntry | string, ...optionalParams: any[]) {
+        this.write("DEBUG", entry, ...optionalParams);
+    }
+
+    warn(entry: LogEntry | string, ...optionalParams: any[]) {
```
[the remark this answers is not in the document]
patuwwy: fatal is actually missing, thx

packages/types/src/object-logger.ts
```diff
@@ -0,0 +1,22 @@
...
+    data?: any[];
+    from: string;
+}>
+
+export interface IObjectLogger {
+    write(level: LogEntry["level"], entry: LogEntry | string, ...optionalParams: any[]): void;
+    trace(entry: LogEntry | string, ...optionalParams: any[]): void;
+    debug(entry: LogEntry | string, ...optionalParams: any[]): void;
+    info(entry: LogEntry | string): void
+    pipe(target: Writable | IObjectLogger): void;
```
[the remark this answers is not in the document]
patuwwy: yes , and fatal as well

packages/obj-logger/src/obj-logger.ts
```diff
@@ -0,0 +1,121 @@
+/* eslint-disable no-console */
+import { DataStream, StringStream } from "scramjet";
+import { PassThrough, Writable } from "stream";
+import { IObjectLogger, LogEntry, LogLevel } from "@scramjet/types";
+import { getName } from "./utils/get-name";
+
+export class ObjLogger implements IObjectLogger {
+    outLogStream = new PassThrough({ objectMode: true });
+    inLogStream = new PassThrough({ objectMode: true });
+    inStringStream = new PassThrough({ objectMode: false });
```
alicja-gruzdz, line 10: I think we chatted about it but it escaped me, can you please remind why it was introduced?

alicja-gruzdz (approved):

packages/host/src/lib/host.ts
patuwwy: get rid. `/log` now provides logs in nd-json format

packages/obj-logger/src/obj-logger.ts
patuwwy: i will add comments to this class :)

patuwwy:
> ```
> objLogger: IObjectLogger;
> ```
> 
> is used and others where it's like:
> 
> ```
> objLogger: ObjLogger;
> ```
> 
> Minor thing but seems inconsistent 🤔

done

packages/adapters/src/dockerode-docker-helper.ts
```diff
@@ -184,14 +184,14 @@ export class DockerodeDockerHelper implements IDockerHelper {
...
-            this.logger.log("Start pulling image", name);
+            this.objLogger.trace("Start pulling image", name);
 
             const pullStream = await this.dockerode.pull(name);
 
             // Wait for pull to finish
             await new Promise(res => this.dockerode.modem.followProgress(pullStream, res));
 
-            this.logger.log("Pulling image", name, "done.");
+            this.objLogger.trace("Image pulled");
```
MichalCz, line 194:
```suggestion
            this.objLogger.trace("Image pulled", name);
```

packages/obj-logger/src/obj-logger.ts
```diff
@@ -0,0 +1,172 @@
...
+    constructor(reference: any, baseLog: LogEntry = {}, logLevel: LogLevel = "TRACE") {
+        this.name = getName(reference);
+
+        this.baseLog = baseLog;
+        this.logLevel = logLevel;
+
+        StringStream
+            .from(this.inputStringifiedLogStream)
+            .JSONParse(true)
+            .pipe(this.inputLogStream);
```
MichalCz, line 67: Do we need to parse this every time? We could just mux our log into the others. It's ok to leave it like this, but maybe a tech debt entry would be in order?

packages/obj-logger/src/utils/pretty-print.ts
```diff
@@ -0,0 +1,17 @@
...
+    DEBUG: COLORS.FgBlue,
+    INFO: COLORS.FgCyan,
+    WARN: COLORS.FgYellow,
+    ERROR: COLORS.FgRed,
+    FATAL: COLORS.FgMagenta
+};
+
+const getLevelColor = (level: LogLevel = "INFO") => `${COLOR_MAP[level]}${level?.padEnd(5)}${COLORS.Reset}`;
+
+export const prettyPrint = (obj: LogEntry) => {
```
MichalCz, line 15: Couldn't find any references in the code? Shouldn't this be in cli?

MichalCz, line 15: We should have another version with no color if color was not turned on, or logs are sent to non tty.

MichalCz: https://github.com/chalk/supports-color

Some inspiration...

packages/types/src/sth-configuration.ts
```diff
@@ -68,6 +70,11 @@ export type HostConfig = {
 }
 
 export type STHConfiguration = {
+    /**
+     * Logging level.
+     */
+    logLevel: LogLevel
```
MichalCz, line 76: `logColor` maybe

packages/obj-logger/src/obj-logger.ts
patuwwy: In the present solution, we parse incoming logs to retrieve some of the data from them and complete the data of the logger they come to.

inputStringifiedLogStream is used only when stringified data are sent, for example via http (runner -> instance-adapter)

packages/types/src/sth-configuration.ts
patuwwy: added

f1ames: :thinking:

packages/obj-logger/src/utils/pretty-print.ts
patuwwy: this is used in host https://github.com/scramjetorg/transform-hub/blob/feature/obj-logger/packages/host/src/lib/host.ts#L120

f1ames (approved):
Looking good :+1: :rainbow: 

image

---

Since `objLogger` is now the only logger (in most places) maybe we can call all it's instances `logger` (I see it's already done in some places), I mean in classes like `this.logger`? A minor thing and it would require renaming in all places so consider it a suggestion only :smile:

packages/obj-logger/src/obj-logger.ts
MichalCz: Ok, my comment is not applicable in that case.

patuwwy:
@f1ames all `objLogger` fields renamed to `logger`

PR #213, opened 2022-01-20
Python runner - first part

jan-warchol:
Rewrite python runner PoC and use it in instance-adapter

Old version assumed the communication with the platform will happen over
unix pipes, which is no longer the case (we use TCP connections now).

The mechanism used in process-instance-adapter for choosing which runner
to use is very primitive - just checks file extension of sequence's
entrypoint.

Note that message codes and communication channel numbering is just
copied over from TS files and hardcoded.

Also, this new runner initially prints messages to a temporary log file,
so that they can be accessed before communication with host is
established.

For now, the assumption is that input and output channels will carry
text, so they are automatically decoded/encoded.

Also, absolutely no Docker support for now.


packages/reference-apps/python-alice/alice.py
```diff
@@ -0,0 +1,13 @@
...
+import json
+from scramjet.streams import Stream
+
+async def delayed_hello(person):
+    await asyncio.sleep(0.25)
+    return f"Hello {person['name'].strip()}!"
+
+def run(input):
+    # FIXME relative path doesn't work
+    source = open('/home/jan/src/transform-hub/packages/reference-apps/python-alice/data.json')
```
Tatarinho: use relative path

Tatarinho (commented):
What about adding type hints?

packages/reference-apps/python-s3-demo/christmas.py
```diff
@@ -0,0 +1,36 @@
...
+        (2.0,  f'\nHej {name} :)\n'),
+        (2.0, 'Ta sekwencja jest napisana w Pythonie 🐍'),
+        (1.0, 'i została uruchomiona'),
+        (0.5, 'za pomocą Pythonowego runnera! 😀'),
+        (2.0, 'Jak widać, potrafi czytać z stdin'),
+        (0.5, 'i pisać na output. 💻'),
+        (2.0, 'Na styku sekwencji i runnera'),
+        (0.5, 'jest odrobina pythonowego frameworka. 🚀'),
+        (2.0,  f'\nWesołych Świąt, {name}! 🎁\n'),
+        (3.0,  f'PS oczywiście na razie wszystko jest zrobione na taśmę klejącą 🩹 😉\n'),
```
MichalCz, line 22: Can we translate this to English?

packages/runner-python/runner.py
```diff
@@ -0,0 +1,117 @@
+import asyncio
+import sys
+import os
+import codecs
+import importlib.util
+from scramjet.streams import Stream
+from hardcoded_magic_values import CommunicationChannels as CC
+from hardcoded_magic_values import RunnerMessageCodes as msg_codes
```
f1ames, line 8: Love such explicit naming :grin:

alicja-gruzdz (commented):
Cool, Python Runner is born! 

It would be helpful to have in-code comments or even docstrings.

I'm leaving just one comment because we've already reviewed the code :)

alicja-gruzdz:
> What about adding type hints?

+1 and Optional

alicja-gruzdz (approved):
:clap:

pietrzakacper (approved):
I don't speak Pythonish, but since this is not a PR to that release branch I can give my blessing.
Would love to go through the code with you next week :wink:

packages/reference-apps/python-s3-demo/christmas.py
jan-warchol: Done.

packages/reference-apps/python-alice/alice.py
jan-warchol: Fixed.

jan-warchol:
> > What about adding type hints?
> 
> +1 and Optional

Yeah, that's definitely something that should be done. The only question is what's the priority.

Tatarinho (approved):

MichalCz (approved):

jan-warchol:
I rewrote the branch to include scramjet framework, and move python stuff to separate dir, as agreed.

PR #218, opened 2022-01-24
K8s adapter PoC

pietrzakacper:


f1ames:
AFAIU after we had a talk yesterday, what we would like to do is:

1. When sequence is uploaded, extract and identify it in a Host process (the same way it is done in `ProcessRunner` via its Sequence Adapter).
2. Start the Runner (this is managed by Instance Adapter?).
3. Copy sequence files (extracted or `.tar.gz`?) and start it on runner using `kubectl` magic (`[copy](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#cp) + [exec](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#exec)`).

Below is a visualization or more-or-less what should be the flow:

image

Some context what was the main problem here (and how it compares to what we have in `Docker` and `Process` runners) would be very useful both now and for the future reference. I don't have enough context for now to describe it, so @pietrzakacper if you would like to do it, it will be great :+1:

PR #219, opened 2022-01-24
Support rapid building for development using esbuild

pietrzakacper:
Implemented development building with `esbuild`.
We are able to build `sth` and `runner` JS bundles and use them in BDD tests, blazingly fast :rocket: 
The only inconvenience is that we have to spawn runner with `ESBUILD=1` env var to point on different Runner package.
(This could probably be easily improved, though it requires time for debugging that I don't want to spend right now :wink:) 
I also had to migrate our typescript source code to use `--esModuleInterop` (`import fs as 'fs'` instead of `import * as fs from 'fs'`) since this what esbuild requires. (This is something that we would have to probably do in near future, since some dependencies do not support any other option)

### Verify 
```bash
git clean -xdf
yarn install && yarn build:refapps && yarn packseq 
yarn build:packages-esbuild
ESBUILD=1 NO_DOCKER=1 yarn test:bdd --name="PT-004 TC-001"
```

### What's next?
Guided by the 80/20 principle I would say the rapid local iterations is the most important leverage of this technology.
However I investigated what should we do if we wanted to tottally migrate for esbuild and it includes:

* [ ] Adjusting Dockerfile builds to new bundle-based build structure
* [ ] Supporting npm package installation (`main` and `bin` entries in package.json would have to change probably)
* [ ] Probably ditching support for `ts-node` since it would be faster just to build
* [ ] Adjusting CI build and start scripts to use new build structure
* [ ] Adjusting BDD tests to spawn using new structure
* [ ] Building types separately for npm release using `tsc --emitDeclarationOnly`
* [ ] Build types on CI using `tsc --noEmit`
* [ ] Implement scripts in each package for building/watching types correctness `tsc --noEmit --watch` while developing

We might want to come back to these, when we would like to focus on decreasing CI execution time.

### DIFF
--- a/ENV_VARS.md
+++ b/ENV_VARS.md
@@ -8,6 +8,10 @@
 
 `SCRAMJET_DEVELOPMENT: boolean`
 
+---
+`ESBUILD: boolean`
+After STH is built with esbuild it should run with this flag
+
 ### BDD tests
 
 `SCRAMJET_SPAWN_JS: boolean (default: false)`
--- a/bdd/lib/utils.ts
+++ b/bdd/lib/utils.ts
@@ -1,5 +1,5 @@
 /* eslint-disable no-loop-func */
-import * as fs from "fs";
+import fs from "fs";
 import { strict as assert } from "assert";
 import { promisify } from "util";
 import { exec, spawn } from "child_process";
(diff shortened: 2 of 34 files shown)


a-tylenda (approved):
verified:
image

alicja-gruzdz (approved):
:fast_forward:

alicja-gruzdz (commented):
neat, could you please add ESBUILD to our ENV Bible? (ENV_VARS.md)
thanks!

PR #222, opened 2022-01-25
Fix: CLI last id memory

MichalCz:
This is a bugfix for the id replacement in CLI.

CLI currently wrongly supplies `sequence id` for `instance` endpoints. This small fix corrects it.


alicja-gruzdz (approved):

a-tylenda (approved):
verified:

image

PR #223, opened 2022-01-25
Add support for complex packages

Tatarinho:


Tatarinho:
I wonder if it's safe to add every package directory to sys.path without cleaning it when the sequence is terminated?

MichalCz (approved):

jan-warchol:
> I wonder if it's safe to add every package directory to sys.path without cleaning it when the sequence is terminated?

Aren't we adding it only for the lifetime of the sequence? I'm pretty sure setting sys.path is limited in scope to current process, and each sequence is ran with a separate runner process.

jan-warchol (approved):

PR #224, opened 2022-01-25
Py runner: better chunk size handling for input and control streams

jan-warchol:
See commit messages for description.

### DIFF
--- a/python/reference-apps/debug-input/debug-input.py
+++ b/python/reference-apps/debug-input/debug-input.py
@@ -1,0 +1,20 @@
+import asyncio
+from scramjet.streams import Stream
+
+# Forward input to output, truncating long chunks and additionally
+# printing to stdout for easier inspection.
+
+def truncate(data):
+    if len(data) > 60:
+        return f'Got chunk: {data[:12]!r} ... {data[-12:]!r} (len: {len(data)})'
+    else:
+        return f'Got chunk: {data!r}'
+
+def run(context, input):
+    return (
+        Stream
+            .read_from(input, max_parallel=1)
+            .each(lambda x: asyncio.sleep(0.5))
+            .map(truncate)
+            .each(print)
+    )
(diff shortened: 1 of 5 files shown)


Tatarinho (approved):

jan-warchol:
I've added a fix to https://app.clickup.com/t/24308805/SCP-2486 - it's related and small, so I think it makes sense to have it in one PR.

Tatarinho (approved):

MichalCz (approved):

python/reference-apps/debug-input/debug-input.py
```diff
@@ -0,0 +1,11 @@
+import asyncio
+import json
+from scramjet.streams import Stream
+
+async def delayed_echo(data):
```
alicja-gruzdz, line 5: We lack reference-apps comments also in the Node version, but I think it would be super useful for us later to add even just one comment line to explain why this app was created (we struggle with searching through them in the Node apps)

alicja-gruzdz (approved):

jan-warchol:
Instructions for verifying first fix:
```
si pack ./python/reference-apps/debug-input -o debug-input.tar.gz
si seq send debug-input.tar.gz
# send file as text - chunks should correspond to lines
si seq start -
si inst input - README.md
# send file as binary - all chunks except last one should have the same length
si seq start -
si inst input -t 'application/octet-stream' - README.md
```

python/runner/runner.py
```diff
@@ -92,15 +96,23 @@ async def handshake(self):
...
             return data['appConfig'], data['args']
 
 
     async def connect_control_stream(self):
-        async for bytes in self.streams[CC.CONTROL]:
-            code, data = json.loads(bytes.decode())
+        # Control stream carries ndjson, so it's enough to split into lines.
+        control_messages = (
+            Stream
+                .read_from(self.streams[CC.CONTROL], chunk_size=65536)
```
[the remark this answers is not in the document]
jan-warchol: Updated the value with a comment :)

python/reference-apps/debug-input/debug-input.py
jan-warchol: Added a comment.

alicja-gruzdz: thanks!

PR #227, opened 2022-01-27
Change of runner selection

Tatarinho:


alicja-gruzdz (approved):

MichalCz (approved):

jan-warchol (approved):
Looks good, only two minor things.

packages/adapters/src/process-sequence-adapter.ts
```diff
@@ -27,9 +27,11 @@ async function getRunnerConfigForStoredSequence(sequencesRoot: string, id: strin
     const packageJson = await readStreamedJSON(createReadStream(packageJsonPath));
 
     const validPackageJson = await sequencePackageJSONDecoder.decodeToPromise(packageJson);
+    const engines = validPackageJson.engines ? { ...validPackageJson.engines } : {};
```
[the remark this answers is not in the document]
Tatarinho: here i copy pasted exactly same behaviour as we have for docker-sequence-adapter, i just assumed its safe and tested

MichalCz (approved):

packages/adapters/src/process-instance-adapter.ts
```diff
@@ -49,12 +49,20 @@ IComponent {
...
-            return [
-                "python3",
-                path.resolve(__dirname, "../../../python/runner/runner.py")
-            ];
+            if (engines.length > 1) {
+                throw new Error("Incorrect config pased to SequenceConfig," +
+                "'engines' field can't contain more than one element");
+            }
+
+            if (engines[0] === "python3") {
```
patuwwy, line 59: ```if ("python3" in config.engines) {````
or 
```if (~engines.indexof("python3") {```
of 
```if (engines.includes("pyhon3") {```

coz, "python3" could be on any place in keys if there is more than 1 required engine specified.

**Review by patuwwy — CHANGES_REQUESTED** [review]
one comment from me :)

File: packages/adapters/src/process-instance-adapter.ts (hunk 2 above)
Comment (Tatarinho) [reply packages/adapters/src/process-instance-adapter.ts]: if (engines.length > 1)  solve this assumption for now, but I've changed as you requested, it's more safe solution for future :)

File: packages/adapters/src/process-instance-adapter.ts [hunk 3]
```diff
@@ -48,9 +48,15 @@ IComponent {
     }
 
     getRunnerCmd(config: SequenceConfig) {
-        if (config.entrypointPath.endsWith(".py")) {
-            this.logger.trace(gotPython);
+        const engines = Object.keys(config.engines);
+
+        if (engines.length > 1) {
+            throw new Error("Incorrect config pased to SequenceConfig," +
```
patuwwy, line 54:
```suggestion
            throw new Error("Incorrect config passed to SequenceConfig," +
```

patuwwy (approved):

PR #232, opened 2022-01-30
Log error w/msg when Logger can't parse incoming logs

patuwwy:
ObjLogger will log error message when incoming string can't be parsed

### DIFF
--- a/packages/types/src/object-logger.ts
+++ b/packages/types/src/object-logger.ts
@@ -1,19 +1,47 @@
-import { DataStream } from "scramjet";
 import { PassThrough, Writable } from "stream";
-import { DeepPartial } from "./utils";
+
+import { DataStream } from "scramjet";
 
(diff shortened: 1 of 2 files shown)


packages/types/src/object-logger.ts
```diff
@@ -1,19 +1,47 @@
...
+    /**
+     * Timestamp. Unix time.
+     */
     ts: number;
-    data?: any[];
+
+    /**
+     * Additional log data.
+     */
+    data: any[];
```
jan-warchol, line 40: Do I understand correctly that from now on the `data` attribute must be present in all log entries?

jan-warchol (approved):
change looks good, just one clarification requested

packages/types/src/object-logger.ts
patuwwy: no, the type is ```Partial<>``` so every field is optional

alicja-gruzdz (approved):

PR #233, opened 2022-01-31
added ssl support in api-server

kociolekscramjet:
Extension for api-server config which makes sslKeyPath and sslCertPath as possible option properties.

The server property has priority over sslKeyPath and sslCertPath. If server is not provided and sslKeyPath and sslCertPath exists then https server instance is created and passed as cero server config property.


packages/api-server/src/index.ts
```diff
@@ -45,6 +47,25 @@ function safeDecorator(cb: (req: IncomingMessage) => MaybePromise<void>) {
...
+        const sslConfig = {
+            key: fs.readFileSync(conf.sslKeyPath),
+            cert: fs.readFileSync(conf.sslCertPath)
+        };
+
+        return https.createServer(sslConfig);
+    }
+
+    return undefined;
+}
```
patuwwy, line 67:
```suggestion
function createCeroServerConfig(conf: ServerConfig = {}): ServerConfig["server"] {
    if (conf.server) {
        return conf.server;
    }

    if (conf.sslKeyPath && conf.sslCertPath) {
        const sslConfig = {
            key: fs.readFileSync(conf.sslKeyPath),
            cert: fs.readFileSync(conf.sslCertPath)
        };

        return createHttpsServer(sslConfig);
    }

    return createHttpServer();
}
```

packages/api-server/src/index.ts
```diff
@@ -11,6 +11,8 @@ import { CeroRouter, CeroRouterConfig } from "./lib/definitions";
 export type ServerConfig = {
     verbose?: boolean;
     server?: Server;
```
patuwwy, line 13:
```suggestion
    server?: HttpsServer | HttpServer;
```

patuwwy: requires
```
import { IncomingMessage, ServerResponse, Server as HttpServer, createServer as createHttpServer } from "http";
import { Server as HttpsServer, createServer as createHttpsServer } from "https";
```

packages/api-server/src/index.ts
patuwwy: what do you think?

kociolekscramjet: Brilliant! Didn't know how to approach that type thanks! :)

alicja-gruzdz (approved):
Nice we have a support for both! 2 in 1!

MichalCz (approved):

PR #236, opened 2022-02-01
Add flag to docker adapter to set RO/RW mode for volume

patuwwy:
DockerSequenceAdapter mounts volume in RW mode
DockerIntanceAdapter mounts volume in RO mode

Setting dirs access permissions moved to Docker Pre-Runner

### DIFF
--- a/packages/adapters/src/types.ts
+++ b/packages/adapters/src/types.ts
@@ -47,7 +47,7 @@ export type DockerAdapterVolumeConfig = {
         volume: DockerVolume;
     } | {
         /**
-         * @property {string} bind a bind mount.
+         * @property {string} bind A bind mount.
          */
         bind: string;
     }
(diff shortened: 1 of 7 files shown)


packages/adapters/src/types.ts
```diff
@@ -30,13 +30,26 @@ export type DockerVolume = string;
...
     {
-        /** @property {DockerVolume} volume Volume. */ volume: DockerVolume
+        /**
+         * @property {DockerVolume} volume Volume.
+         */
+        volume: DockerVolume;
     } | {
-        /** @property {DockerVolume} volume Volume. */ bind: string
+        /**
+         * @property {DockerVolume} volume Volume.
```
alicja-gruzdz, line 50:
```suggestion
         * @property {string} bind a bind mount.
```

alicja-gruzdz (approved):
Thanks :raised_hands:

patuwwy:
Need to move setting permission from Runner to PreRunner

PR #239, opened 2022-02-02
IaC startup mode

MichalCz:
In order for STH to be compatible with general Infrastructure as Code approach and larger stacks, it will need to be able to start with a predefined set of services.

This PR aims to implement such mode with `ProcessSequenceAdapter` in such a way that STH could be bundled with sequences and packaged in a container.

This is Work in Progress still.


MichalCz:
@pietrzakacper I added DTO's - this is a start, but this goes in a similar way to what you were thinking.

MichalCz:
It seems the test procedure does result in an error.

patuwwy (approved):
lg

packages/utility/src/typeguards/dto/sequence-start.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function isStartSequenceDTO(arg: any): arg is StartSequenceDTO {
+    if (typeof arg !== "object") throw new Error("DTO is not an object");
+    const { id, appConfig, args, ...rest } = arg;
+
+    if (typeof id !== "string") throw new Error("DTO id is not string");
+    if (!["object", "undefined"].includes(typeof appConfig)) throw new Error(`DTO appConfig is ${typeof appConfig}, not an object`);
+    if (typeof args !== "undefined") {
+        if (!Array.isArray(args)) throw new Error("DTO args are not an array");
+        if ((args as string[]).some(x => typeof x !== "string")) throw new Error("DTO args are all stings");
```
alicja-gruzdz, line 11:
```suggestion
        if ((args as string[]).some(x => typeof x !== "string")) throw new Error("DTO args are all strings");
```

alicja-gruzdz (approved):
I ran a couple of our reference apps with the startup config following the given format and it worked nicely.

The `--exit-with-last-instance` is a cool option.

alicja-gruzdz:
Running with `endless-names-output` refapp with and without exit with last  Instance:
[endless_exit_with_instance.txt](https://github.com/scramjetorg/transform-hub/files/8534653/endless_exit_with_instance.txt)
[endless_without_instance_exit.txt](https://github.com/scramjetorg/transform-hub/files/8534654/endless_without_instance_exit.txt)

config:
```
{
  "sequences": [
    { "id": "endless-names-output", "args": ["5"], "appConfig": {} }
  ]
}
```

PR #240, opened 2022-02-03
Sending PANG during handshake

Tatarinho:
Support for the PANG messages in the Runner for the default Sequences which do NOT require or provide any topic.
The full topic support will be implemented in another PR.


alicja-gruzdz (approved):

MichalCz (approved):

MichalCz:
Let's merge.

Issue #206, opened 2022-01-14
Starting a failed sequence hangs indefinitely.


MichalCz:
**Describe the bug**
When a sequence is started, for instance with `si seq start sequenceId` and the sequence exits immediately (throws an error) the CLI process hangs and waits forever.

**Steps to Reproduce**

1. Create a sequence that fails (has an error in syntax)
2. Run: `si seq pack <dir> && si seq send <tgz> && si seq start <id>` - see that the start fails.

***Reproduction repo/branch***

Working on it.

**Expected behavior**

An error should be thrown and exitcode 1 should be sent.

**Version (please complete the following information):**
 - STH version: 0.14.0
 - node version: 16.x
 - os: Ubuntu Server 20.04

**Additional information like logs, screenshots etc.**

```
Starting host on port: 34859
Spawning with command: node ../dist/sth/bin/hub.js -P 34859 --instances-server-port 38877 --no-docker
2022-01-14T23:31:35.944Z log (object:Host) Host main called.
2022-01-14T23:31:35.948Z info (object:SocketServer) Server on: { address: '::', family: 'IPv6', port: 38877 }
2022-01-14T23:31:35.951Z info (object:Host) API listening on: 0.0.0.0:34859
Feature: CLI tests # features/e2e/E2E-010-cli.feature:1

  @ci
  Scenario: E2E-010 TC-022 Check minus replacements # features/e2e/E2E-010-cli.feature:203
    Given host is running
2022-01-14T23:31:35.983Z Starting request to http://localhost:34859/api/v1/load-check
2022-01-14T23:31:36.003Z Request ok: http://localhost:34859/api/v1/load-check status: 200 OK
    When I execute CLI with "pack ../packages/reference-apps/hello-alice-out" arguments
2022-01-14T23:31:36.003Z debug (object:Host) Request date: 2022-01-14T23:31:35.987Z, method: GET, url: /api/v1/load-check, status: 200
Spawning command /usr/bin/env si pack ../packages/reference-apps/hello-alice-out -a http://localhost:34859/api/v1
    And I execute CLI with "seq send -" arguments
Spawning command /usr/bin/env si seq send - -a http://localhost:34859/api/v1
2022-01-14T23:31:40.534Z info (object:Host) New sequence incoming...
2022-01-14T23:31:40.545Z info (object:Host) Sequence identified: {
  type: 'process',
  entrypointPath: 'index',
  version: '0.14.0',
  name: '@scramjet/hello-alice-out',
[...]

PR #244, opened 2022-02-04
Do not start instance from broken sequence

patuwwy:
WIP

Added sequences to `packages/reference-apps`:
- [x] `no-index`
- [x] `syntax-error`
- [x] `throw-error`

- [x] empty-seq
- [x] topic-only

- [ ] Fix starting / reject starting above sequences
-----

This PR is going to fix starting instance from broken sequences in several cases and is a response for #206


packages/host/src/lib/csi-controller.ts
```diff
@@ -137,19 +137,23 @@ export class CSIController extends TypedEmitter<Events> {
...
+
             this.startInstance();
         });
 
-        i.then(() => this.main()).catch(e => {
-            this.emit("error", e);
-        });
+        i.then(() => this.main())
+            .catch(e => {
+                this.logger.error("Instance failed to start", e);
```
patuwwy, line 155: there is no listener for "error" event

PR #247, opened 2022-02-08
feat: Kubernetes Runtime Adapters

pietrzakacper:
### Verify
1. Setup deployment on cluster:https://github.com/scramjetorg/docs/tree/docs/th-k8s/developers-instructions/transform-hub-k8s.
  2. And then run BDD tests, connecting to STH pod 
 ```
 LOCAL_HOST_BASE_URL=http://<sth-svc-external-ip>/api/v1 SCRAMJET_HOST_BASE_URL=http://<sth-svc-external-ip>:8000/api/v1 RUNTIME_ADAPTER=kubernetes NO_HOST=1 yarn test:bdd-ci
 ```

### Description

1. When sequence is uploaded, extract and identify it in a Host process (the same way it is done in `ProcessSequenceAdapter`).
2. Start the Runner with the runner process waiting to start before uploading
3. Copy sequence files and start it on runner using `kubectl` [exec](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#exec)`).

Below is a visualization or more-or-less what should be the flow:

image

### TODO `All Green`
* [x] Rebase [KP]
* [x] Add k8s namespace to STH configuration [KK]
* [x] Shut down Runner Pod when Runner process ends (maybe there's a way to do that automatically) [KP]
* [x] Implment KubernetesInstanceAdapter kill [KK]
* [x] Wait for Runner's Pod to enter Running state before executing any commands [KK]
* [x] Fix troublesome tests: [KP]
      * ~E2E-006 TC-001~
      * ~E2E-001 TC-002~
      * ~PT-002 TC-001~
      * ~E2E-010 TC-017~
* [x] Implement Get Runner PID in BDD tests (maybe noop implementation for this iteration is fine :wink:)
* [x] Fix CLI tests [KK]
    * Requires setting `LOCAL_HOST_BASE_URL` to STH instance address.
    * Fails on deleting sequence (entire Host process errors due to [rethrown not catched error](https://github.com/scramjetorg/transform-hub/blob/e8afaaf9026193913c6b064dd225dc212a95927d/packages/host/src/lib/host.ts#L368))
* [x] Deleting sequence (at least via CLI) errors 
* [x] Run BDD tests locally on Kubernetes cluster

### TODO `Becoming Production Ready` / `Review` ready
* [x] Add all the cluster specific information (such as namespace, sth service name, runner image) to STH configuration [KK]
* [x] Readme to kubernetes deployment scripts - see https://github.com/scramjetorg/docs/pull/96
* [x] Implement proper adapter selection mechanism (possibly `--adapter="k8s" | "docker" | "process"`) (eridacate no docker) 
* [x] Generate Runner pod name based on ID
* [x] Implement Kubernetes client Adapter

### Follow-ups
* [ ] ~**Rewrite prerunner so it would prepare image for runner**~ - Extracted as https://github.com/scramjetorg/transform-hub/issues/267
* [ ] ~Unpack only package.json (ideally to stdout) in KubernetesSequenceAdapter~
* [ ] ~Store only compressed package on disk, such as `.scramjet_sequences/<seq-id>.tar.gz`~
* [ ] ~Figure out a way to get rid of redundant scripts `unpack.sh` and `run.sh` from Runner docker image~
* [ ] ~Support [topic feature tests](https://github.com/scramjetorg/transform-hub/blob/bae0c49bc9783bf8ce41aaccbea0369358c2a971/bdd/features/e2e/E2E-013-topic.feature)~
* [ ] ~Investigate objLogger not printing in KubernetesInstanceAdapter~
* [ ] ~Extract objLogger improvements to a separate PR~
* [ ] ~Make yarn clean use `git clean -xdf` or also remove packed sequences~

All but first item from the above list moved to https://github.com/scramjetorg/transform-hub/issues/268.


f1ames:
CLI tests are now passing. I had to introduce one change though which is kind of unrelated - #255. CI test can be run locally with:

```bash
LOCAL_HOST_BASE_URL=http://sth.url:8000/api/v1 \
SCRAMJET_HOST_BASE_URL=http://sth.url:8000/api/v1 \
K8S=1 \
NO_HOST=1 \
DEVELOPMENT=true \
SCRAMJET_TEST_LOG=1 \
yarn test:bdd-ci --tags="@cli"
```

Still deleting sequence throws an error so it needs to be investigated.

f1ames:
I have checked rest of BDD tests we have (ones which starts host itself). There are some which seems not necessary (related to starting host itself and passing configs) but also some related to topics. Topics related cases are not covered in other tests so it might be good to have coverage on k8s too. And so:

1. HUB tests

https://github.com/scramjetorg/transform-hub/blob/bae0c49bc9783bf8ce41aaccbea0369358c2a971/bdd/features/e2e/hub/HUB-001-host-config.feature#L3-L87

Tests for starting HUB on its own. Not necessary for now IMHO. OTOH, if we assume that one can start HUB with k8s adapter (the same as with `docker` or `process` one) on its own, without our infrastructure it could be good to have some test coverage for that.

2. Topic feature - https://github.com/scramjetorg/transform-hub/blob/bae0c49bc9783bf8ce41aaccbea0369358c2a971/bdd/features/e2e/E2E-013-topic.feature

I think we should have coverage here too. So it will be good to run those tests with k8s adapter.

3. Topic feature (via CLI)

Those are the same as `E2E-013-topic.feature` tests (but done via `CLI`) so it seems not necessary for now too.

https://github.com/scramjetorg/transform-hub/blob/bae0c49bc9783bf8ce41aaccbea0369358c2a971/bdd/features/e2e/E2E-010-cli.feature#L155-L200

pietrzakacper:
There seem to be various connectivity issues while running Transform-Hub on the k8s cluster.

`Runner` sometimes fails to connect with following errors:
```
Error: getaddrinfo ENOTFOUND sth-runner
```
```
Error: getaddrinfo EAI_AGAIN sth-runner
```
both regarding DNS server.

And `Host` sometimes cannot connect to kubeapi server with following error:
```
Error: ENETUNREACH 10.96.0.1:443
```

pietrzakacper:
Additionally, Runner pod sometimes takes more than 20 second to spin up (to switch from Pending to Running state).
Which causes our bdd tests to randomly fail bc of timeout reached on sequence start.

pietrzakacper:
Problems with connectivity on a cluster occur even using kubectl on my virtual machine:
image

f1ames:
For setup and testing instructions please refer to https://github.com/scramjetorg/docs/tree/docs/th-k8s/developers-instructions/transform-hub-k8s.

packages/adapters/src/kubernetes-client-adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+        const kubeApi = this.config.makeApiClient(k8s.CoreV1Api);
+
+        const result = await this.runWithRetries(retries, "Create Pod", async () => {
+            return await kubeApi.createNamespacedPod(this._namespace, {
+                apiVersion: "v1",
+                kind: "Pod",
+                metadata,
+                spec
+            });
+        });
```
MichalCz, line 56: Looks like a nice utility function. Maybe some refactoring should be done.

Also `return await` doesn't allow promises to chain in tail call and I don't really see a benefit here?

```suggestion
        const result = await this.runWithRetries(retries, "Create Pod", async () => 
            kubeApi.createNamespacedPod(this._namespace, {
                apiVersion: "v1",
                kind: "Pod",
                metadata,
                spec
            }));
```

packages/adapters/src/kubernetes-client-adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+        while (true) {
+            try {
+                const response = await kubeApi.readNamespacedPodStatus(podName, this._namespace);
+                const status = response.body.status?.phase || "";
+
+                if (expectedStatuses.includes(status)) {
+                    return status;
+                }
+            } catch (err: any) {
+                this.logger.error("Failed to get pod status", err);
```
MichalCz, line 101: Won't this fall into an infinite loop and just spam the system with log message twice every second?

packages/adapters/src/kubernetes-instance-adapter.ts
```diff
@@ -0,0 +1,178 @@
...
+} from "@scramjet/types";
+
+import path from "path";
+import { ObjLogger } from "@scramjet/obj-logger";
+import { createReadStream } from "fs";
+import { KubernetesClientAdapter } from "./kubernetes-client-adapter";
+
+import { JsonDecoder } from "ts.data.json";
+
+const adapterConfigDecoder = JsonDecoder.object<K8SAdapterConfiguration>({
```
MichalCz, line 21: Nice typeguard. We should have this in the config. Let's create a backlog entry out of this.

packages/sth/src/bin/hub.ts
```diff
@@ -11,14 +11,18 @@ const options = program
     .option("-E, --identify-existing", "Index existing volumes as sequences", false)
     .option("-C, --cpm-url <host:ip>")
     .option("-I, --id <id>")
-    .option("--no-docker", "Run all the instances on the host machine instead of in docker containers. UNSAFE FOR RUNNING ARBITRARY CODE.", false)
-    .option("--sequences-root", "Only works with --no-docker option. Where should ProcessSequenceAdapter save new sequences")
+    .option("--runtime-adapter <adapter>", "Determines adapters used for loading and starting sequence. One of 'docker', 'process', 'kubernetes'", "docker")
```
MichalCz, line 14:
```suggestion
    .option("--runtime-adapter <adapter>", "Determines adapters used for loading and starting sequence. One of 'docker', 'process', 'kubernetes'", "docker")
    .option("--no-docker", "Run all the instances on the host machine instead of in docker containers. UNSAFE FOR RUNNING ARBITRARY CODE.", false)
```

I'd still leave this as a shorthand so that we won't break backwards compatibility.

packages/sth/src/bin/hub.ts
```diff
@@ -11,14 +11,18 @@ const options = program
...
     .option("--runner-image <image name>", "Image used by runner")
     .option("--runner-max-mem <mb>", "Maximum mem used by runner")
     .option("--prerunner-image <image name>", "Image used by prerunner")
     .option("--prerunner-max-mem <mb>", "Maximum mem used by prerunner")
     .option("--expose-host-ip <ip>", "Host IP address that the Runner container's port is mapped to.")
     .option("--isp, --instances-server-port <port>", "Port on which server that instances connect to should run.")
+    .option("--sequences-root <path>", "Only works with --runtime-adapter='process' option. Where should ProcessSequenceAdapter save new sequences")
```
MichalCz, line 21: Kube would need a separate sequences root, since it stores the data in a different format.

MichalCz (changes requested):
Please fix the backwards compatibility and use separate roots for tgz and directories in kube and process adapters respectively.

daro1337:
We should add example with POD manifest and  minimal RBAC to README.md for users

packages/adapters/src/kubernetes-instance-adapter.ts
pietrzakacper: https://github.com/scramjetorg/transform-hub/issues/279

packages/adapters/src/kubernetes-client-adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+
+                this._config = kc;
+            } catch (err: any) {
+                this.logger.error("Unable to load kubeconfig", err);
+            }
+        }
+
+        if (!this._config) {
+            throw new Error("Kubernetes API client not initialized");
+        }
```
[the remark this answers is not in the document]
pietrzakacper: fixed

packages/adapters/src/kubernetes-client-adapter.ts
pietrzakacper: fixed

packages/adapters/src/kubernetes-client-adapter.ts
```diff
@@ -0,0 +1,140 @@
...
+                const status = response.body.status?.phase || "";
+
+                if (expectedStatuses.includes(status)) {
+                    return status;
+                }
+            } catch (err: any) {
+                this.logger.error("Failed to get pod status", err);
+            }
+
+            await defer(500);
```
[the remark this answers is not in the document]
pietrzakacper: fixed

packages/adapters/src/kubernetes-client-adapter.ts
pietrzakacper: fixed

packages/adapters/src/kubernetes-instance-adapter.ts
```diff
@@ -0,0 +1,178 @@
...
+        const env = [
+            `SEQUENCE_PATH=${path.join("/package", config.entrypointPath)}`,
+            `DEVELOPMENT=${process.env.DEVELOPMENT ?? ""}`,
+            `PRODUCTION=${process.env.PRODUCTION ?? ""}`,
+            `INSTANCES_SERVER_PORT=${instancesServerPort}`,
+            `INSTANCES_SERVER_HOST=${this.adapterConfig.sthPodHost}`,
+            `INSTANCE_ID=${instanceId}`,
+        ]
+            .map(str => str.split("="))
+            .map(([name, value]) => ({ name, value }));
```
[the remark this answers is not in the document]
pietrzakacper: Improved it slightly. It was like that bc I copied it from other adapters that required string encoded structure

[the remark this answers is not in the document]
pietrzakacper: not needed now

packages/adapters/src/kubernetes-sequence-adapter.ts
```diff
@@ -0,0 +1,130 @@
...
+        this.logger = new ObjLogger(this);
+    }
+
+    /**
+     * Initializes adapter.
+     *
+     * @returns {Promise<void>} Promise resolving after initialization.
+     */
+    async init(): Promise<void> {
+        return fs.access(this.config.sequencesRoot)
```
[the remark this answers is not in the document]
pietrzakacper: NOW it is :wink:

packages/sth/src/bin/hub.ts
pietrzakacper: fixed

packages/sth/src/bin/hub.ts
pietrzakacper: fixed

MichalCz (approved):

f1ames:
Docs on testing kubernetes adapter with minikube are now complete, tested and can be seen here - https://github.com/scramjetorg/docs/blob/main/developers-instructions/transform-hub-k8s/README.md.

packages/reference-apps/stdio-sequence/index.ts
```diff
@@ -14,6 +14,8 @@ module.exports = async function(_stream: any) {
         .from(process.stdin)
         .lines("\n")
         .parse((str: any) => [+(str.match(/^\w+/) || []).pop(), str])
+        /// THIS IS SUPER WEIRRRRRRRDDDDD
```
pietrzakacper, line 17: Cleanup

packages/adapters/src/kubernetes-instance-adapter.ts
```diff
@@ -0,0 +1,171 @@
...
+                }
+            },
+            {
+                containers: [{
+                    env,
+                    name: runnerName,
+                    image: this.adapterConfig.runnerImage,
+                    stdin: true,
+                    command: ["wait-for-sequence-and-start.sh"],
+                    imagePullPolicy: "Always"
```
pietrzakacper, line 101: We should remove that or hide it under DEVELOPMENT env var

f1ames: Added to https://github.com/scramjetorg/transform-hub/issues/268.

PR #253, opened 2022-02-09
Support for graceful stop in Python Runner

Tatarinho:


Tatarinho:
si inst stop <id> <timeout>

alicja-gruzdz (commented):
cool, thanks. The timeout and the SEQUENCE_STOPPED messages work as expected. 
I'm not giving an approve just yet, because I've talked with @jan-warchol. He said we already have [context](https://github.com/scramjetorg/transform-hub/blob/release/0.16/packages/runner/src/runner-app-context.ts#L70) working in Python Runner, so we can call the `stopHandler` method on the context, like it is in the node version https://github.com/scramjetorg/transform-hub/blob/release/0.16/packages/runner/src/runner.ts#L240
We can talk about it tomorrow.

alicja-gruzdz (approved):

PR #254, opened 2022-02-09
Add healthy monitoring sequence for Python Runner

Tatarinho:


alicja-gruzdz (commented):
Is it the sample for this BDD test? https://github.com/scramjetorg/transform-hub/blob/release/0.16/bdd/features/e2e/E2E-005-monitoring.feature#L13

alicja-gruzdz:
I've checked it's https://github.com/scramjetorg/transform-hub/blob/release/0.16/packages/reference-apps/healthy-sequence/index.ts. It looks good. We should probably start thinking about organizing the Python reference-apps directory and moving the samples there.

alicja-gruzdz (approved):

PR #255, opened 2022-02-10
Return error response on delete seq error

f1ames:
I'm not 100% sure if that would be a valid change regarding how we would like to handle such errors but in the current state it results in crashing entire host process which I assume we would like to avoid:

> Error response is returned instead of rethrowing an error. The error is thrown in the top level layer (host) and there is nothing to catch it. This results in crashing entire host process.

### DIFF
--- a/packages/host/src/lib/host.ts
+++ b/packages/host/src/lib/host.ts
@@ -365,7 +365,10 @@ export class Host implements IComponent {
         } catch (error: any) {
             this.logger.error("Error removing sequence!", error);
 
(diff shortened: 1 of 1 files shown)


alicja-gruzdz (approved):
That's major, thanks for fixing.
We need to look at `main` and other methods and make sure we catch them all.

MichalCz (approved):

f1ames:
Rebased onto latest `release/0.16`.

PR #256, opened 2022-02-10
Get rid of long lines in BDD code

jan-warchol:
* Shorten some very long step names in Cucumber
* Format BDD step definitions to get rid of extremely long lines

Formated with

      prettier --write --print-width 90 --tab-width 4


alicja-gruzdz (approved):

Tatarinho (approved):

MichalCz:
We cannot accept this change now. We're working on these files and any non planned changes may hinder our progress.

PR #266, opened 2022-02-16
Long run sequence with generator

ErykSol:


packages/reference-apps/csv-transform/index.js
```diff
@@ -0,0 +1,37 @@
...
+const { createHash } = require("crypto");
+const { createInterface } = require("readline");
+
+const app = async function (input) {
+    let outputStream = new PassThrough({ encoding: "utf-8" });
+    const rl = createInterface({ input });
+    let drain = null;
+
+    rl.on("line", async (line) => {
+        const out = outputStream.write(line + ";" + createHash("md5").update(line).digest("hex") + "\n");
```
alicja-gruzdz, line 11: Generally, the Sequence code looks solid and elegant.
If we add getting the magic data from the object store like we spoke ( https://github.com/scramjetorg/transform-hub/blob/release/0.16/bdd/features/performance-tests/PT-003-process-large-file.feature#L16 ) then we're home.

MichalCz (approved):

packages/reference-apps/csv-transform/index.js
ErykSol: Adding data from store finished

alicja-gruzdz (approved):
super, thanks!

PR #269, opened 2022-02-16
Support for topics in Python Runner

Tatarinho:
**`yarn install:python-deps`** before

To run bdd tests:
**RUNTIME_ADAPTER=process yarn test:bdd --tags=@python**

1. Run sth
`DEVELOPMENT=1 yarn start:dev --no-docker`
2. pack seq and send to sth
3. To test consumer, send some data on topic:
`si topic send pytopic-test python/reference-apps/python-alice/data.json`
4. Data can be retrieved also with
  `si topic get pytopic-test`

### DIFF
--- a/python/runner/runner.py
+++ b/python/runner/runner.py
@@ -6,6 +6,7 @@ import json
 from pyee import EventEmitter
 import importlib.util
 from io import DEFAULT_BUFFER_SIZE as CHUNK_SIZE
+
 from scramjet.streams import Stream
 from logging_setup import LoggingSetup
 from hardcoded_magic_values import CommunicationChannels as CC
@@ -186,11 +187,15 @@ class Runner:
 
         self.logger.info(f'Sending PANG')
         monitoring = self.streams[CC.MONITORING]
-        pang_provides_data = {
-            'provides': '',
-            'contentType': ''
-        }
-        send_encoded_msg(monitoring, msg_codes.PANG, pang_provides_data)
+
+        if hasattr(self.sequence, 'provides'):
+            produces = self.sequence.provides
+            self.logger.info(f'Sending PANG with {produces}')
+            send_encoded_msg(monitoring, msg_codes.PANG, produces)
+        if hasattr(self.sequence, 'requires'):
+            consumes = self.sequence.requires
+            self.logger.info(f'Sending PANG with {consumes}')
+            send_encoded_msg(monitoring, msg_codes.PANG, consumes)
 
         if asyncio.iscoroutine(result):
             result = await result
@@ -203,14 +208,17 @@ class Runner:
 
 
     async def connect_input_stream(self, input_stream):
-        raw_headers = await self.streams[CC.IN].readuntil(b'\r\n\r\n')
-        header_list = raw_headers.decode().rstrip().split('\r\n')
-        headers = {
-            key.lower(): val for key, val in [el.split(': ') for el in header_list]
-        }
-        self.logger.info(f'Input headers: {repr(headers)}')
+        if hasattr(self.sequence, "requires"):
+            input_type = self.sequence.requires['contentType']
+        else:
+            raw_headers = await self.streams[CC.IN].readuntil(b'\r\n\r\n')
+            header_list = raw_headers.decode().rstrip().split('\r\n')
+            headers = {
+                key.lower(): val for key, val in [el.split(': ') for el in header_list]
+            }
+            self.logger.info(f'Input headers: {repr(headers)}')
+            input_type = headers['content-type']
 
-        input_type = headers['content-type']
         if input_type == 'text/plain':
             input = Stream.read_from(self.streams[CC.IN])
             self.logger.debug('Decoding input stream...')
--- a/bdd/README.md
+++ b/bdd/README.md
@@ -96,7 +96,7 @@ for seq_name in `ls python/reference-apps/`; do si pack "python/reference-apps/$
 Also, python runner does not support Docker yet, so run tests with
 
 ```bash
-NO_DOCKER=1 yarn test:bdd --tags=@python
+RUNTIME_ADAPTER=process yarn test:bdd --tags=@python
 ```
 
 ### Results :bar_chart:
--- a/bdd/features/e2e/E2E-014-python.feature
+++ b/bdd/features/e2e/E2E-014-python.feature
@@ -92,7 +92,7 @@ Feature: Test our shiny new Python runner
         Then instance health is "false"
         And host is still running
 
-    @ignore
+
     @python
     Scenario: E2E-014 TC-011 Send data between python instances using topics
         Given host is running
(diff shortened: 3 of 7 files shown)


MichalCz (approved):

python/runner/runner.py
```diff
@@ -177,63 +178,76 @@ async def run_instance(self, config, args):
...
+        if (not hasattr(self.sequence, 'topic')):
+            raw_headers = await self.streams[CC.IN].readuntil(b'\r\n\r\n')
+            header_list = raw_headers.decode().rstrip().split('\r\n')
+            headers = {
+                key.lower(): val for key, val in [el.split(': ') for el in header_list]
+            }
+            self.logger.info(f'Input headers: {repr(headers)}')
+            input_type = headers['content-type']
+        else:
+            input_type = 'text/plain'
```
jan-warchol, line 222: I think content type should not be hardcoded.

python/runner/runner.py
```diff
@@ -177,63 +178,76 @@ async def run_instance(self, config, args):
...
 
 
     async def connect_input_stream(self, input_stream):
-        raw_headers = await self.streams[CC.IN].readuntil(b"\r\n\r\n")
-        header_list = raw_headers.decode().rstrip().split("\r\n")
-        headers = {
-            key.lower(): val for key, val in [el.split(": ") for el in header_list]
-        }
-        self.logger.info(f"Input headers: {repr(headers)}")
+        if (not hasattr(self.sequence, 'topic')):
```
jan-warchol, line 213: I understand that when input comes from a topic it doesn't have headers, but is that expected or is it a bug? Does it always behave in this way? I think we need a confirmation from Ala or someone else more familiar with how sth works.

python/reference-apps/topics/main.py
```diff
@@ -0,0 +1,5 @@
+def topic():
```
jan-warchol, line 1: could it be a variable instead of a function? I think having a function that just returns a value is a bit confusing.

jan-warchol (changes requested):
Does this pass the BDD test for topics in python runner (https://github.com/scramjetorg/transform-hub/blob/release/0.16/bdd/features/e2e/E2E-014-python.feature#L97)? I think support for both incoming ("requires") and outgoing ("provides") topics is necessary.

Also, please separate formatting changes from actual logic changes, at least by putting them in a separate commit. Otherwise it's hard to find the important parts of the diff.

python/runner/runner.py
Tatarinho: @alicja-gruzdz what do you think?

alicja-gruzdz: The topics currently only support `application/x-ndjson`. However, in the later stages we should support other content types.
I checked by running the Sequences that require and provide the same topic:
https://github.com/scramjetorg/transform-hub/tree/release/0.16/packages/reference-apps/delay-out-in-a
https://github.com/scramjetorg/transform-hub/tree/release/0.16/packages/reference-apps/delay-out-in-b
and the information about content type came in the headers to `delay-out-in-b`, which requires the topic:
Zrzut ekranu z 2022-02-21 10-06-59

python/runner/runner.py
Tatarinho: There is no way to get contentType from the host, I'm getting it from the sequence itself now

python/runner/runner.py
Tatarinho: I talked with @pietrzakacper and he confirmed that there is no way to get headers in this case, in tsrunner headers are retrieved from the sequence

alicja-gruzdz: Sorry, you're right that in the case of `requires` it sets https://github.com/scramjetorg/transform-hub/blob/release/0.17/packages/runner/src/runner.ts#L336

pietrzakacper: Yes, we have two sepeare use cases for determining content type of the input stream:
1. User writes to sequence input through REST API (or CLI on top of it). In this case the[ input stream will receive  headers](https://github.com/scramjetorg/transform-hub/blob/release/0.17/packages/runner/src/runner.ts#L351) containing `Content-Type`.
2. Sequence requires a topic as an input stream. In this case [content-type is specified by the Sequence](https://github.com/scramjetorg/transform-hub/blob/release/0.17/packages/runner/src/runner.ts#L336) (`contentType` field next to `requires`). No headers will be sent on the input stream, just the data that sits on the topic.

python/runner/runner.py
```diff
@@ -186,11 +187,16 @@ async def run_instance(self, config, args):
...
-        pang_provides_data = {
+        pang_data = {
             'provides': '',
             'contentType': ''
         }
-        send_encoded_msg(monitoring, msg_codes.PANG, pang_provides_data)
+        send_encoded_msg(monitoring, msg_codes.PANG, pang_data)
+
+        if hasattr(self.sequence, 'METADATA'):
+            metadata = self.sequence.METADATA
```
MichalCz, line 197: I'm a bit iffy on this - I'd rather see `produces` and `consumes`... Metadata seems to mean nothing specific.

MichalCz (changes requested):
I'm not sure about the naming convention, take a look at the comment there.

python/runner/runner.py
```diff
@@ -186,11 +187,20 @@ async def run_instance(self, config, args):
 
         self.logger.info(f'Sending PANG')
         monitoring = self.streams[CC.MONITORING]
-        pang_provides_data = {
+        pang_data = {
             'provides': '',
             'contentType': ''
         }
-        send_encoded_msg(monitoring, msg_codes.PANG, pang_provides_data)
+        send_encoded_msg(monitoring, msg_codes.PANG, pang_data)
```
jan-warchol, line 194: Do we need to send empty PANG at all? This looks like a leftover from previous version of the code ;)

python/runner/runner.py
```diff
@@ -203,14 +213,19 @@ async def run_instance(self, config, args):
...
 
     async def connect_input_stream(self, input_stream):
-        raw_headers = await self.streams[CC.IN].readuntil(b'\r\n\r\n')
-        header_list = raw_headers.decode().rstrip().split('\r\n')
-        headers = {
-            key.lower(): val for key, val in [el.split(': ') for el in header_list]
-        }
-        self.logger.info(f'Input headers: {repr(headers)}')
+        if hasattr(self.sequence, 'PROVIDES'):
+            input_type = self.sequence.PROVIDES['contentType']
```
jan-warchol, line 217: I don't think we should set _input_type_ based on `PROVIDES` ;p

jan-warchol (changes requested):
Almost there! Just found a couple minor things.
Oh, please don't forget to un-ignore the BDD test case (right now it's disabled)

jan-warchol (approved):
I've checked and it works!

alicja-gruzdz (approved):

MichalCz (approved):

alicja-gruzdz:
`RUNTIME_ADAPTER=process yarn test:bdd --tags=@python`

[verify.txt](https://github.com/scramjetorg/transform-hub/files/8132539/verify.txt)

PR #275, opened 2022-02-17
Py runner: implement event handling, passing BDD test

jan-warchol:
This PR depends on #271 and includes a commit from that branch.


Tatarinho (approved):

python/reference-apps/python-events/events.py
```diff
@@ -0,0 +1,8 @@
+import asyncio
+
+async def run(context, input):
+    def respond_to_event(message):
+        context.emit('test-response', f"reply to {message}")
```
Tatarinho, line 5: I would change "" to ''

Tatarinho (approved):
Looks good to me, one tiny change requested

MichalCz (approved):

python/reference-apps/python-events/events.py
jan-warchol: done

PR #286, opened 2022-02-23
Improve 'upgrade:all' script

f1ames:
I have reworked script for aligning and updating all dependencies in all packages in the project. It works similar to what we had before with a few changes. I also tried to make it as simple as possible (well, almost).

## How to run or test it:

From the main project directory, run:

```bash
yarn upgrade:all
```

I already run it and push updated deps as separate PR - https://github.com/scramjetorg/transform-hub/pull/287.

## How does it work:

1. It gathers all the dependencies and their versions from all the `package.json` files. Which files to choose is done based on [hardcoded glob path](https://github.com/scramjetorg/transform-hub/blob/16ef9512e0faa0ac8190e200d43b89911b70100f/scripts/deps-update.js#L39).
2. For each dependency it stores all the unique versions which were found. We don't need this ATM, but if we would like to have aligning deps to highest/lowest version, that will be useful.
3. It chooses highest version for each dependency and creates `all-deps/package.json` and `all-deps/package.previous.json` files based on that (those files are the same at this point).
4. The it runs `npm install --package-lock-only` in `all-deps/` directory to generate `package-lock.json` file based on created `package.json`.
5. Based on `package-lock.json` file, updated packages versions are extracted and written to `package.json` file (see bloew on version extraction details).
6. Last step is just comparing `all-deps/package.json` to `all-deps/package.previous.json` and printing out which packages were updated.
7. The script is integrated into `yarn upgrade:all` command which after running it, propagates updated deps to all the packages (via `yarn align`), removes `all-deps` directory and runs `yarn install`.

### Version extraction

As for the 5th point. package version from `package-lock.json` can be obtained in two ways. Based on `packages...dependencies.name` or based on `dependencies.name.version` (see below). In the first case it would be unchanged and the same as in `package.json` file. In the second case it is the latest version which would be installed and that's the version which is used.

Case 1:

```json
{
  "packages": {
    "": {
      "name": "@scramjet/all-deps",
      "version": "^0.17.0",
      "dependencies": {
        "@kubernetes/client-node": "^0.16.2",
```

Case 2:

```json
{
  "dependencies": {
    "@kubernetes/client-node": {
      "version": "0.16.3",
```

## Sample output:

image
image
image

## Further improvements

The scripts now does everything at once. If we would like to have aligning only (without upgrading), slight modifications would be needed.


MichalCz (approved):

MichalCz:
One question though... the update removed all the caret signs in front of local deps, like this for instance

```
    "@scramjet/types": "^0.17.0",
```

Only in `dependencies`, not `devDependencies`.

Is that in any way rational?

f1ames:
> One question though... the update removed all the caret signs in front of local deps, like this for instance
> 
> ```
>     "@scramjet/types": "^0.17.0",
> ```
> 
> Only in `dependencies`, not `devDependencies`.
> 
> Is that in any way rational?

Hmmm, this should not happen I guess. I'll have to investigate this :mag:

MichalCz:
Ok, let's fix this in this PR and do deps upgrade before every release.

f1ames:
> Ok, let's fix this in this PR and do deps upgrade before every release.

@MichalCz I have already covered this in PR which update deps, here - https://github.com/scramjetorg/transform-hub/pull/287. Is that ok for you?

MichalCz:
Let's merge both of those

f1ames:
Rebased onto latest `release/0.17`.

PR #290, opened 2022-02-24
Force reinstalling python dependencies

jan-warchol:
We discovered that if (for some reason) the installation was broken,
running yarn install:python-deps would not fix it.


Tatarinho (approved):

alicja-gruzdz (approved):
super, thanks!

Issue #258, opened 2022-02-11
Runner is not killed by InstanceAdapter when not responding to KILL/STOP control msg


pietrzakacper:
**Describe the bug**
Technically everything works correctly bc our Runner is made in such a way that it always politely quits when asked.
Practically part of our code in CSIController will never be executed leading to  stale containers/processes if runner doesn't exit on its own.

The bug was introduced after absorbing supervisor to CSIController. Specifically it's all the control stream handlers that won't be executed, such as:
https://github.com/scramjetorg/transform-hub/blob/release/0.16/packages/host/src/lib/csi-controller.ts#L285
https://github.com/scramjetorg/transform-hub/blob/release/0.16/packages/host/src/lib/csi-controller.ts#L305

What will happen is the REST API (`_kill`, `_stop`) will pipe the control message to the control ***downstream*** and the `communicationHandler` will only dispatch the control messages to handlers from ***upstream***

The fix would be to intercept REST API request, run the handlers in CSIControllers and the send it on control stream

patuwwy:
Probably related to #206, #244

patuwwy:
Especially to https://github.com/scramjetorg/transform-hub/issues/206#issuecomment-1030082183

PR #294, opened 2022-02-25
Restore proper STOP/KILL supervisor mechanisms in CSIController

pietrzakacper:
This fixes: #258 

The implemented behavior should resemble the one from before Supervisor removal.
I guess all these flows should be revisited and refactored. They are a bit messy and quite complex.

Changes include:
* Process kill after Runner is not exiting after KILL control msg
* Send KILL msg after Runner is not emitting SEQUENCE_STOPPED after STOP control msg
* Process kill after Runner is not exiting after sending SEQUENCE_COMPLETED
* BDD test to confirm that Runner that is not responding is correctly killed

### Veryify
New CI test should confirm the case when Runner is not responding and the instance adapter removes the process forcefully after instance kill command

### DIFF
--- a/packages/host/src/lib/csi-controller.ts
+++ b/packages/host/src/lib/csi-controller.ts
@@ -38,6 +38,7 @@ import { getRouter } from "@scramjet/api-server";
 import { getInstanceAdapter } from "@scramjet/adapters";
 import { defer, promiseTimeout, TypedEmitter } from "@scramjet/utility";
 import { ObjLogger } from "@scramjet/obj-logger";
+import { ReasonPhrases } from "http-status-codes";
 
 const runnerExitDelay = 11000;
 
@@ -478,10 +479,13 @@ export class CSIController extends TypedEmitter<Events> {
         this.router.op("post", "/_kill", async () => {
             await this.communicationHandler.sendControlMessage(RunnerMessageCode.KILL, {});
 
-            await promiseTimeout(this.endOfSequence, runnerExitDelay)
+            // This will be resolved after HTTP response. It's not awaited on purpose
+            promiseTimeout(this.endOfSequence, runnerExitDelay)
                 .catch(() => this.instanceAdapter.remove());
 
-            return {};
+            return {
+                opStatus: ReasonPhrases.ACCEPTED
+            };
         }, this.communicationHandler);
     }
 
--- a/packages/host/src/lib/csi-controller.ts
+++ b/packages/host/src/lib/csi-controller.ts
@@ -537,7 +537,6 @@ export class CSIController extends TypedEmitter<Events> {
     }
 
     async handleSequenceStopped(message: EncodedMessage<RunnerMessageCode.SEQUENCE_STOPPED>) {
-        // @TODO redesign this process, currently it's not used by any sequence
         this.logger.trace("Sequence ended, sending kill");
 
         try {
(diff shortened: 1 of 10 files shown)


packages/host/src/lib/csi-controller.ts
```diff
@@ -38,8 +38,9 @@ import { getRouter } from "@scramjet/api-server";
 import { getInstanceAdapter } from "@scramjet/adapters";
 import { defer, promiseTimeout, TypedEmitter } from "@scramjet/utility";
 import { ObjLogger } from "@scramjet/obj-logger";
+import { ReasonPhrases } from "http-status-codes";
 
-const stopTimeout = 7000;
+const runnerExitDelay = 11000;
```
jan-warchol, line 43: why 11000? Is it related to the `exitDelay` in `runner.ts`?

packages/runner/src/runner.ts
```diff
@@ -371,14 +371,16 @@ export class Runner<X extends AppConfig> implements IComponent {
         try {
             await this.runSequence(sequence, args);
 
+            this.writeMonitoringMessage([RunnerMessageCode.SEQUENCE_COMPLETED, {}]);
             this.logger.trace(`Sequence completed. Waiting ${exitDelay}ms with exit.`);
 
-            await defer(exitDelay); // @TODO: investigate why we need to wait
+            await defer(exitDelay);
```
jan-warchol, line 377: Do you know why the delay is needed?

pietrzakacper: yes, it is needed to perform API checks after sequence finish

pietrzakacper: mainly for our BDD tests

jan-warchol (approved):
LGTM

packages/host/src/lib/csi-controller.ts
```diff
@@ -467,8 +457,36 @@ export class CSIController extends TypedEmitter<Events> {
         // operations
         this.router.op("post", "/_monitoring_rate", RunnerMessageCode.MONITORING_RATE, this.communicationHandler);
         this.router.op("post", "/_event", RunnerMessageCode.EVENT, this.communicationHandler);
-        this.router.op("post", "/_stop", RunnerMessageCode.STOP, this.communicationHandler);
-        this.router.op("post", "/_kill", RunnerMessageCode.KILL, this.communicationHandler);
+
+        this.router.op("post", "/_stop", async (req) => {
```
MichalCz, line 461: @pietrzakacper can you extract this function to a class method?

packages/host/src/lib/csi-controller.ts
```diff
@@ -467,8 +457,36 @@ export class CSIController extends TypedEmitter<Events> {
...
+            await defer(timeout);
+
+            if (!this.keepAliveRequested) {
+                await this.communicationHandler.sendControlMessage(RunnerMessageCode.KILL, {});
+            }
+
+            return {};
+        }, this.communicationHandler);
+
+        this.router.op("post", "/_kill", async () => {
```
MichalCz, line 479: And here as well...

MichalCz (commented):
@pietrzakacper take a look at the comment...

patuwwy (changes requested):
please rebase and see comments from other reviewers

packages/host/src/lib/csi-controller.ts
pietrzakacper: fixed

packages/host/src/lib/csi-controller.ts
pietrzakacper: fixed

patuwwy (approved):

Issue #295, opened 2022-02-26
incorrect docker sequence adapter error


patuwwy:
Sending sequence in unsupported format...

Container logs:
```
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
{"error":"Invalid pkg tar.gz archive"}
```

DockerSequenceAdapter error:
```
{"level":"ERROR","msg":" decoder failed at key \"main\" with error: undefined is not a valid string","ts":1645884718656,"from":"DockerSequenceAdapter","Host":{},"data":["d2d48be0-6a9a-48c1-96cb-996d8c376f4d"]
```

patuwwy:
Needs to be added to kubernetes adapter as well

PR #296, opened 2022-02-26
Fix not passing error from prerunner

patuwwy:
fix for #295 

before: docker sequence adapter did not handle the error from PreRunner and tried to parse the data from package.json
```
2022-02-26T14:34:56.686Z ERROR DockerSequenceAdapter <SequencePackageJSON> decoder failed at key "main" with error: undefined is not a valid string [ '9a0798a7-92f9-4708-870b-fe44cab30e72' ]
```

with this fix
```
2022-02-26T15:03:48.948Z DEBUG DockerSequenceAdapter PreRunner response [ { error: 'Invalid pkg tar.gz archive' } ]
2022-02-26T15:03:48.948Z ERROR DockerSequenceAdapter PreRunner failed [ 'Invalid pkg tar.gz archive' ]
```

response for `POST /sequence`:
```
HTTP/1.1 422 Unprocessable Entity
...

{"error":{"code":"PRERUNNER_ERROR","data":"Invalid pkg tar.gz archive"}}
```

### DIFF
--- a/packages/adapters/src/docker-sequence-adapter.ts
+++ b/packages/adapters/src/docker-sequence-adapter.ts
@@ -4,11 +4,17 @@ import {
     SequenceConfig,
     STHConfiguration,
     DockerSequenceConfig,
-    IObjectLogger,
+    IObjectLogger
 } from "@scramjet/types";
 import { Readable } from "stream";
 import { DockerodeDockerHelper } from "./dockerode-docker-helper";
-import { DockerAdapterResources, DockerAdapterRunResponse, DockerAdapterStreams, DockerVolume, IDockerHelper } from "./types";
+import {
+    DockerAdapterResources,
+    DockerAdapterRunResponse,
+    DockerAdapterStreams,
+    DockerVolume,
+    IDockerHelper
+} from "./types";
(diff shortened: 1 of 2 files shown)


alicja-gruzdz:
Zrzut ekranu z 2022-03-02 12-32-37

alicja-gruzdz (approved):
In Docker Adapter it works great.
I assume that this PR does not include Process Adapter that needs to be handled separately (currently no meaningful info about the cause of error there - I only checked while I was at it). I'm also not sure about k8s.

alicja-gruzdz:
> This is a huge error log in the right panel. let me try to make it a bit shorter.

Indeed, I didn't want to bother you with it. If you can do something about it, it will be excellent.

patuwwy:
image

patuwwy:
> I assume that this PR does not include Process Adapter that needs to be handled separately (currently no meaningful info about the cause of error there - I only checked while I was at it). I'm also not sure about k8s.

added to Process Adapter as well
no k8s yet

alicja-gruzdz:
> added to Process Adapter as well

It's working great, thanks.
Zrzut ekranu z 2022-03-03 09-17-17

PR #300, opened 2022-03-01
Get rid of waits in HUB tests

f1ames:
This is a simple example of how we can get rid of explicit waits in tests. Here I just optimized one test suite (`HUB`).

Test command:
```
yarn test:bdd-ci-no-host --name="HUB-001"
```

Before:
```
9 scenarios (9 passed)
40 steps (40 passed)
0m28.342s (executing steps: 0m28.268s)
Done in 32.56s.
```

After:
```
9 scenarios (9 passed)
37 steps (37 passed)
0m23.382s (executing steps: 0m23.311s)
Done in 27.66s.
```

So it's 5 seconds less (which seems not to be that much), but in fact it is ~18% of the entire tests duration.

### DIFF
--- a/bdd/step-definitions/hub/config.ts
+++ b/bdd/step-definitions/hub/config.ts
@@ -15,6 +15,8 @@ import { ReadStream } from "fs";
 import { PassThrough } from "stream";
 import { defer } from "../../lib/utils";
 
+const AWAITING_POLL_DEFER_TIME = 250;
+
 When("hub process is started with parameters {string}", function(this: CustomWorld, params: string) {
     return new Promise<void>((resolve, reject) => {
         this.resources.hub = spawn(
@@ -94,7 +96,7 @@ Then("get runner container information", { timeout: 20000 }, async function(this
             this.resources.containerInfo = info;
             this.resources.containerInspect = inspect;
         } else {
-            await defer(50);
+            await defer(AWAITING_POLL_DEFER_TIME);
         }
     }
 });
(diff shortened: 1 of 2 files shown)


bdd/step-definitions/hub/config.ts
```diff
@@ -75,18 +75,28 @@ Then("exit hub process", function(this: CustomWorld) {
...
+                new Dockerode().listContainers().then(
+                    containers => containers.find(container => container.Id === containerId)),
+                new Dockerode().getContainer(containerId!).inspect(),
+            ]);
+
+            this.resources.containerStats = stats;
+            this.resources.containerInfo = info;
+            this.resources.containerInspect = inspect;
+        } else {
+            await defer(50);
```
MichalCz, line 97: I'd up that to `250` it's 5 times less hits, and not much time to add.

bdd/step-definitions/hub/config.ts
```diff
@@ -121,17 +131,29 @@ Then("send fake stream as sequence", async function(this: CustomWorld) {
     );
 });
 
-Then("end fake stream", async function(this: CustomWorld) {
-    this.resources.pkgFake.end();
-    await defer(2000);
+Then("end fake stream", async function(this: CustomWorld): Promise<void> {
+    return new Promise(res => {
+        this.resources.pkgFake.on("close", async () => {
+            await defer(50);
```
MichalCz, line 137: Also let's use a constant somewhere above

```js
const AWAITING_POLL_DEFER_TIME = 250
```

MichalCz (approved):

bdd/step-definitions/hub/config.ts
f1ames: See https://github.com/scramjetorg/transform-hub/pull/300/commits/683aa442dfcbfcf1059d2f28da47a6aa7df26eec.

a-tylenda (approved):

PR #303, opened 2022-03-02
BDD tests: remove unneeded test cases

jan-warchol:
* "E2E-001 TC-001 Execute hello-alice-out example for host" and
  "E2E-001 TC-003 - KM5_Cloud Server Instance Component" duplicate
  "E2E-001 TC-002 API test - Get instance output"
* "E2E-003 TC-002 Kill sequence - kill handler should emit event when
  executed" was ignored for a very long time with a comment "potential
  edge case", so it appears to be no longer relevant
* "E2E-004 TC-002 Send test-event in one function and emit this event in
  another function" was not actually checking what it advertised, so
  it's removed. A ticket for replacing it in the future:
  https://app.clickup.com/t/25d148c
* "Scenario: E2E-005 TC-002 Get monitoring from sequence, should return
  default monitoring value: healthy true" is not needed because we check
  "healthy: true" in most other test cases
* Two HUB scenarios basically duplicated previous two, just with a
  different port parameter.

### DIFF
--- a/bdd/features/e2e/E2E-001-samples.feature
+++ b/bdd/features/e2e/E2E-001-samples.feature
@@ -1,16 +1,5 @@
 Feature: Sample e2e tests
 
-    Scenario: E2E-001 TC-001 Execute hello-alice-out example for host
-        Given host is running
-        When sequence "../packages/reference-apps/hello-alice-out.tar.gz" loaded
-        And instance started
-        And get "output" with instanceId and wait for it to finish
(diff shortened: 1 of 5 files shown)


a-tylenda:
So far so good 😸

f1ames (approved):
Looks good, thanks for explaining why each test was removed :+1:

a-tylenda (approved):

PR #304, opened 2022-03-02
Force-close runners in specific tests

### LINKED ISSUE #193: Decrease time of execution of BDD tests by killing instances
Runner has an exit delay. Some of the tests need to it to be able to get the instances streams even after the sequence has finished processing. After BDD tests perform those queries we can just kill the instances using `kill` endpoint and decrease the time we wait for the process to exit.

f1ames:
Instead of waiting for runner to exit itself, it is now force-closed in `After` step. There were some tests in which checking if runner closed successfully seemed like a part of the test (especially when there was a step sending `kill` to runner instance earlier) so I didn't remove it from such tests.

Tested locally with:

```
yarn test:bdd-ci --name="E2E"
```

Before:

```
29 scenarios (29 passed)
225 steps (225 passed)
5m25.421s (executing steps: 5m24.514s)
```

After:

```
29 scenarios (29 passed)
218 steps (218 passed)
4m47.005s (executing steps: 4m46.117s)
```

**Almost 40 seconds less**.

Closes #193.

### DIFF
--- a/bdd/features/e2e/E2E-002-stop.feature
+++ b/bdd/features/e2e/E2E-002-stop.feature
@@ -1,7 +1,7 @@
 Feature: Stop e2e tests
 
-    @ci @runner-cleanup
-    Scenario: E2E-002 TC-001 API test - Send stop, sequence sends keepAlive
+    @ci
+    Scenario: E2E-002 TC-001 API test - Send stop, sequence sends keepAlive, runner closes successfully
         Given host is running
         When sequence "../packages/reference-apps/can-keep-alive.tar.gz" loaded
         And instance started with arguments "SEND_KEEPALIVE"
@@ -12,6 +12,7 @@ Feature: Stop e2e tests
         And send stop message to instance with arguments timeout 5000 and canCallKeepAlive "true"
         And wait for instance healthy is "true"
         And send stop message to instance with arguments timeout 0 and canCallKeepAlive "false"
+        And runner has ended execution
         Then host is still running
 
     @ci @runner-cleanup
--- a/bdd/features/e2e/E2E-002-stop.feature
+++ b/bdd/features/e2e/E2E-002-stop.feature
@@ -1,6 +1,6 @@
 Feature: Stop e2e tests
 
-    @ci
+    @ci @runner-cleanup
     Scenario: E2E-002 TC-001 API test - Send stop, sequence sends keepAlive
         Given host is running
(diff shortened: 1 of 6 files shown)


alicja-gruzdz (approved):
Nice. It's a good idea and I like the implementation with After hook and tag.

Only one note, I'm not sure we should include all tests related to checking if the stop request works correctly (E2E-002 TC-002, E2E-002 TC-003, E2E-002 TC-001). Maybe at least one of them should stay as is. If I'm not mistaken they test if the Runner process exits after a given timespan. Unless this is covered by some other tests.

patuwwy (approved):
Awesome!

f1ames:
> Only one note, I'm not sure we should include all tests related to checking if the stop request works correctly (E2E-002 TC-002, E2E-002 TC-003, E2E-002 TC-001). Maybe at least one of them should stay as is. If I'm not mistaken they test if the Runner process exits after a given timespan. Unless this is covered by some other tests.

Yes, I was only able to get a context and the reasoning behind those tests based on test name/description and its steps. So that might be a valid point. OTOH, all tests are described with `... - Send stop, sequence sends keepAlive` and this is checked by `And send stop message to instance with arguments timeout...` step which is before waiting for runner. So my assumption is that at this point, the test checked what it should and waiting for runner is irrelevant. But if you recall that it was important to check for runner exiting correctly we can bring it back to some or all mentioned tests and rename them so it's clear what we want to check. WDYT @alicja-gruzdz?

alicja-gruzdz:
> So my assumption is that at this point, the test checked what it should and waiting for runner is irrelevant. But if you recall that it was important to check for runner exiting correctly we can bring it back to some or all mentioned tests and rename them so it's clear what we want to check. WDYT @alicja-gruzdz?


I think your proposition is reasonable. I know our poor naming and the lack of comments are making the tests refactoring task more difficult for you guys. Better naming would be much needed. We could leave one of these tests - `E2E-002 TC-001`. This test encapsulates the most important aspects of stop. Runner should exit if the timeout given in the stop request passed and `canKeepAlive` parameter was false. In general, with the stop request, I think we can't only rely on the Instance Adapter in the Host to terminate the Runner, because the CSI Controller is not aware of the stop request execution. However, please correct me if I'm wrong. But coming back to main question you asked: for me leaving out only `E2E-002 TC-001` (not to be forcefully killed) will be fine.
Great speeding up work.

PR #313, opened 2022-03-03
Run release tests with parallel option

f1ames:
This PR adds `--parallel 2` flag to "release tests" run on CI. I also had to get rid of "global" `let`s in CLI test steps (it still needs a proper cleanup but I wasn't focusing on that :see_no_evil:). After few runs it seems like tests are stable.

OTOH, I tried to do the same with "dev tests" and there was always something failing, kind of randomly. So it looks like we need to clean up and rework those tests a bit to make it run with `--parallel` option.

**Good news** - release tests instead of usual 7 to 9 minutes:

image

now are done in around 5 minutes:

image

**Bad news** - due to [issue with mixed up logs](https://github.com/cucumber/cucumber-js-pretty-formatter/issues/13), for parallel run I changed formatter to `progress` which means it might be harder to conclude something (when test fails) looking on CI output, because there will be no steps (or if enabled, it will be mixed up).

### DIFF
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -49,8 +49,11 @@ jobs:
       - name: Start sth
         run: scramjet-transform-hub >> scramjet.log 2>&1 &
 
-      - name: Run BDD npm tests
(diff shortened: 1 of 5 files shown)


MichalCz:
`21 successful checks`, so let's merge?

MichalCz (approved):

f1ames:
Rebased onto latest `release/0.18`.

PR #327, opened 2022-03-04
Test communicationHandler cleanup

Tatarinho:


packages/host/src/lib/csi-controller.ts
```diff
@@ -220,6 +220,14 @@ export class CSIController extends TypedEmitter<Events> {
         this.instancePromise = instanceMain();
     }
 
+    async cleanup() {
+        await Promise.all([
+            this.instanceAdapter.cleanup(),
+            this.communicationHandler.cleanup()
+        ]);
+        this.logger.error("Cleanup done...");
```
patuwwy, line 228:
```suggestion
        this.logger.info("Cleanup done.");
```

alicja-gruzdz:
Added [`do not merge`](https://github.com/scramjetorg/transform-hub/labels/do%20not%20merge), because we need to add a couple of improvements in the CSI Controller log handling (@Tatarinho knows about it, I'm just leaving it here to let to know not to include it yet in the incoming release)

PR #331, opened 2022-03-09
Add HTTPS support to Verser

pietrzakacper:
Changes include:
* Testing `Verser` for a standard HTTP flow 
* Testing `Verser` for a new HTTPS flow
* Git ignoring encryption secrets
* Tweaking `Verser` to support HTTPS
* Cleaning up CPMConnector from useless `attachServer` and `apiServer`

TODO:
* [ ] actually enable using https in Host and Manager (in a separate PR)

### DIFF
--- a/packages/verser/test/http-connection.spec.ts
+++ b/packages/verser/test/http-connection.spec.ts
@@ -1,0 +1,134 @@
+import test from "ava";
+import * as http from "http";
+import * as https from "https";
+import { createServer } from "@scramjet/api-server";
+import { Verser, VerserClient, VerserConnection } from "../src";
+import path from "path";
+import { readFileSync } from "fs";
+import { APIExpose } from "@scramjet/types";
+import { spawnSync } from "child_process";
+
+async function connectVerserClientAToVerserB(
+    apiB: APIExpose,
+    verserClientA: VerserClient,
+): Promise<VerserConnection> {
+    const verserB = new Verser(apiB.server);
+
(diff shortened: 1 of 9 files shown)


packages/verser/test/http-connection.spec.ts
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable no-console */
+import { defer } from "@scramjet/utility";
+import test from "ava";
+import { request, Server } from "http";
+import { createServer } from "@scramjet/api-server";
+import { Verser, VerserClient, VerserConnection } from "../src";
+
+test("Connect VerserClient A to Verser B and send HTTP GET Request to VerserClient A", async (t) => {
```
MichalCz, line 8: Perhaps you could extract the main portion of the test to a separate function and provide the server from outside, so you could test it with `http2`, `https` and `http`?

pietrzakacper: definitely a good idea

MichalCz (commented):
Regarding local keys and certs, I'd prefer running keygen before tests and adding those to gitignore. Someone could use them just because they're distributed with the package

packages/verser/test/http-connection.spec.ts
pietrzakacper: i think there's too many use case specific details (we not only use http/https servers but also clients) in each test to completely extract it. I cleaned it up a bit by extracting some common steps to util functions.

patuwwy (approved):

MichalCz (approved):

PR #340, opened 2022-03-11
Removing redundant STOP step from host-steps

Tatarinho:


alicja-gruzdz (approved):
Thanks for removing the redundant step :raised_hands: 
We can also have in the stop test step: `and canCallKeepAlive "false"`, because we don't check it this test if `canCallKeepAlive` works.

Tatarinho:
> 

You are right. I've changed that.

PR #342, opened 2022-03-13
Add hub and space commands to cli

patuwwy:


packages/cli/src/bin/index.ts
```diff
@@ -64,6 +64,7 @@ const errorHandler = (err: ClientError) => {
         .option("-L, --log", "Logs all API requests in detail", conf.log)
         .option("-a, --api-url <url>", "Specify base API url", conf.apiUrl)
         .option("-f, --format <value>", "Specify display formatting: json or pretty", conf.format)
+        .option("-f, --middlewareApiUrl <value>", "Specify middleware API url", conf.middlewareApiUrl)
```
MichalCz, line 67: This should not be camel case.

MichalCz: We should consider a new name for api-url too... With a fallback for backward compatibility, but preferably not documented

ErykSol (approved):

alicja-gruzdz (commented):
cool :cloud: cli :sunglasses: 

I've started to verify it. Question, because we will be writing the descriptions of the commands. In the mode: `development`, should it not attempt to send request to the MW at all or (as I understand now is) just modify the Host client?
Zrzut ekranu z 2022-03-15 17-44-01

alicja-gruzdz (commented):
Minor thing but just leaving it here as I came across it. When the space is not set, the `hub list` throws such maybe not so nice looking looking error:
Zrzut ekranu z 2022-03-15 17-48-40

As visible above, when the space is set, the `hub list` works as a charm.
I think we can leave it for now, not important.

alicja-gruzdz (approved):
Sweet!
I set the hub and started the Sequence :heavy_check_mark: 
Zrzut ekranu z 2022-03-15 17-53-19

Zrzut ekranu z 2022-03-15 17-54-46

:clap: :clap: :clap:

PR #344, opened 2022-03-14
Scope config

ErykSol:
Changes in PR:
1) implements part of CLI scope commands:
-scope list
-scope print
-scope use
-scope delete
2) Paths used in project are moved to separate file for reusability reasons.
3) Changes control of config used (between scope config and global config) depending on use case


patuwwy (approved):

alicja-gruzdz (approved):
Thanks, much anticipated sessions coming :)
I can't wait to try it when we have add support in Manager (spaces) and Hubs.

PR #346, opened 2022-03-15
Add support for requirements.txt in Python packages

Tatarinho:
As the part of this PR:
- Support for install all needed deps from requirements.txt file
- Removing Python-Framework from TransformHub repository

**Warning**
Python-Framework needs fix for setup.py to avoid downloading all packages from repository, e.g 'tests' directory 
Depends on: https://github.com/scramjetorg/framework-python/pull/35

**IMPORTANT UPDATE**
Further investigation pushed me in a different direction


alicja-gruzdz (commented):
Looks like it installs the deps fine. I will try to run something later as I don't have anything ready.
Zrzut ekranu z 2022-03-15 21-14-04

MichalCz (approved):

alicja-gruzdz:
@Tatarinho After your investigations, I also tried such approach (screen shot attached) but it didn't work. And I see the issue you explained to me that numpy installs other deps in the local user directory. Well done on checking the source of the Sequence error. Sorry, I couldn't help to solve the issue.


Zrzut ekranu z 2022-03-16 21-49-55

Tatarinho:
I would need more time to investigate issue.

alicja-gruzdz:
> I would need more time to investigate issue.

thanks. 
I talked to my colleague who is an experienced ML Python engineer. He said that [it is a typical installation](https://numpy.org/doc/stable/user/building.html) for a data science library - C code with a wrapper. So it means in the future we would need to compile it for different platforms. Since there are only a handful of such data science libraries maybe for later stages of the project we should consider their preinstallment, at least in the case of Docker.

Tatarinho:
@MichalCz @alicja-gruzdz @jan-warchol
Possibility to have proper way to install extra deps for Python sequences is crucial to show ability of our platform, however it seems that it is more complex feature.

We can't just pack libs into /dist directory. Some of them can works, but a lot of future possible deps, e.g. numpy, sckit-learn and more, won't work, cause they have more dependencies etc.

In this PR, I created simple change, which allows to install all needed deps when sequence is run on ProcessAdapter. Process is 'local Python' right? So me as Python Developer can assume that I should install all deps to run sequence. Better approach would be to run Python from isolated environment, but generally it is the same ProcessAdapter. This changes allow to do it automatically, but it's not solution for Docker/Kubernates. This solution is also very unfriendly for bigger projects, cause it install all dependencies on every "si seq send", so time needed between send and start can be long, and we need to face here possibility of duplicated libraries.

So as I explained, more complicated are Docker/Kubernates adapters. 
My first though was to install them "on the fly" as soon as we send package, but again, multiple directories with same libraries for different sequences and...it will be installed everytime we RUN sequence, on Python Runner container. And of course, as soon as container will be stopped/recreated etc., we need to install all dependencies again. Very ugly solution.

Potential fix for that is to inform developers that they can put their "requirements.txt" into "requirements.txt" from PythonRunner. This will allow to have all libs preinstalled in one place, but...not every possible user wants to build image from scratch etc. I can assume that a lot of developers just need to have "platform" to run sequences. So they won't download source of TransforHub to make some adjustments.

Summary:
1) For now Python environment is not very functional, possible architecture changes?
1) We need to install all deps from requirements.txt of sequence to image used locally by user without his knowledge of Dockerfile etc.
2) We could add possibility to build python-runner image based on "user image with all needed deps", for more advanced users

Tatarinho:
Adding sequenceDir to PYTHONPATH resolved most issues related to that feature

MichalCz (approved):

PR #348, opened 2022-03-15
Introduce serverless host (WIP)

f1ames:
**WIP**


f1ames:
Closing this PR for now as it's outdated and we decided to postpone this feature for later.

My key takeaway here is that separate host classes for server and serverless Hosts might be a bit over-engineered since the actual differences in the code are really minor. And the Host does not create server instance by itself but receives it from the `start-host` script so it could be just optional. OTOH if we decide to split Host into subclasses, the Host itself should be responsible for starting (or not) it's API server.

PR #351, opened 2022-03-16
Set auth token for all si platform requests

patuwwy:
`si auth set-token <token>` to set platform token

### DIFF
--- a/packages/cli/src/lib/commands/auth.ts
+++ b/packages/cli/src/lib/commands/auth.ts
@@ -1,0 +1,25 @@
+import { CommandDefinition } from "../../types";
+import { setConfigValue } from "../config";
+
+/**
+ * Initializes `auth` command.
+ *
+ * @param {Command} program Commander object.
+ */
+export const auth: CommandDefinition = (program) => {
+    const authCmd = program
+        .command("auth [command]");
+
+    /**
+    * Command `si host version`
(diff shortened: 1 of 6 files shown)


alicja-gruzdz (approved):

packages/cli/src/lib/commands/auth.ts
```diff
@@ -0,0 +1,25 @@
...
+    * Command `si host version`
+    * Get STH version number from package.json file
+    * Log: version number
+    */
+    authCmd
+        .command("set-token")
+        .argument("<token>")
+        .description("Set Platform auth token")
+        .action(async (token: string) => {
+            setConfigValue("token", token);
```
ErykSol, line 23: shouldn't we add simple format validation for JWT before save to config?

ErykSol (commented):
nice and clean :)

packages/cli/src/lib/commands/auth.ts
patuwwy: we can add simple API endpoint to platfrom to validate token or just check here if lenght is correct

ErykSol: I would go for simple format check- lenght and number of dots in string

ErykSol (approved):

PR #355, opened 2022-03-17
Feature/commander completion

kociolekscramjet:
Commander competion with bash and install commands

DOD:
- [x] Can see completion script in terminal after `si completion bash` command
- [x] Completion script visible in ~/.bashrc after `si completion install` command
- [x] After installation and `source ~/.bashrc` completion works eg `si comp<tab>` gives `si completion` and `si <tab><tab>` gives possible commands.


patuwwy (approved):

alicja-gruzdz (approved):
Excellent
Zrzut ekranu z 2022-03-18 12-04-50

agne6ka (approved):
✔✔✔✔

PR #357, opened 2022-03-21
Encrypt all Host <-> Manager communication with TLS

pietrzakacper:
This PR includes changes required for encrypting `host` <-> `manager` communication.
Should be reviewed together with: https://github.com/scramjetorg/scramjet-cpm-dev/pull/184

***Verify***: There is a BDD test in the CPM PR: [E2E-002 TC-006](https://github.com/scramjetorg/scramjet-cpm-dev/blob/7f8edf9f5a1ed16b27dce6874bacb7568ba86a5d/bdd/features/e2e/E2E-002-cpm-sd.feature#L58)

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -14,7 +14,8 @@
     "test": "echo no tests yet -- # npm run test:ava",
     "test:ava": "ava",
     "preinstall": "tar --version 2>&1 >/dev/null",
-    "prepack": "node ../../scripts/publish.js"
+    "prepack": "node ../../scripts/publish.js",
+    "postbuild": "yarn prepack"
   },
   "author": "Scramjet <open-source@scramjet.org>",
   "license": "AGPL-3.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -14,8 +14,7 @@
     "test": "echo no tests yet -- # npm run test:ava",
     "test:ava": "ava",
     "preinstall": "tar --version 2>&1 >/dev/null",
-    "prepack": "node ../../scripts/publish.js",
-    "postbuild": "yarn prepack && cp ./src/completion ../../dist/cli/completion -rL && cp ./src/types/commander-completion.d.ts ../../dist/cli/types/commander-completion.d.ts -L"
+    "prepack": "node ../../scripts/publish.js"
   },
   "author": "Scramjet <open-source@scramjet.org>",
   "license": "AGPL-3.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -15,7 +15,7 @@
     "test:ava": "ava",
     "preinstall": "tar --version 2>&1 >/dev/null",
     "prepack": "node ../../scripts/publish.js",
-    "postbuild": "yarn prepack && cp ./src/completion ../../dist/cli/completion -rL && cp ./src/types/commander-completion.d.ts ../../dist/cli/types/commander-completion.d.ts -L"
+    "postbuild": "yarn prepack && mkdir -p ../../dist/cli/types && cp ./src/completion ../../dist/cli/completion -rL && cp ./src/types/commander-completion.d.ts ../../dist/cli/types/commander-completion.d.ts -L"
   },
   "author": "Scramjet <open-source@scramjet.org>",
   "license": "AGPL-3.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -15,7 +15,7 @@
     "test:ava": "ava",
     "preinstall": "tar --version 2>&1 >/dev/null",
     "prepack": "node ../../scripts/publish.js",
-    "postbuild": "yarn prepack"
+    "postbuild": "yarn prepack && cp ./src/completion ../../dist/cli/completion -rL && cp ./src/types/commander-completion.d.ts ../../dist/cli/types/commander-completion.d.ts -L"
   },
   "author": "Scramjet <open-source@scramjet.org>",
   "license": "AGPL-3.0",
--- a/packages/client-utils/src/index.ts
+++ b/packages/client-utils/src/index.ts
@@ -1,6 +1,22 @@
 import { ClientUtilsBase } from "./client-utils";
 import { HttpClientNode as HttpClient, SendStreamOptions } from "./types";
 import * as nodefetch from "node-fetch";
+import http from "http";
+import https from "https";
+
+const nodeFetchWithHttps = (ca?: string | Buffer) =>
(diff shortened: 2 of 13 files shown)


packages/client-utils/src/index.ts
```diff
@@ -1,6 +1,22 @@
 import { ClientUtilsBase } from "./client-utils";
 import { HttpClientNode as HttpClient, SendStreamOptions } from "./types";
 import * as nodefetch from "node-fetch";
+import http from "http";
+import https from "https";
+
+const nodeFetchWithHttps = (ca?: string | Buffer) =>
```
pietrzakacper, line 7: Possibly reuse this function

packages/host/src/lib/cpm-connector.ts
```diff
@@ -539,12 +546,13 @@ export class CPMConnector extends TypedEmitter<Events> {
      * @param topicCfg Topic configuration.
      */
     sendTopic(topic: string, topicCfg: { contentType: string, stream: ReadableStream<any> | WritableStream<any> }) {
-        const cpmUrl = new URL("http://" + this.cpmURL + "/topic/" + topic);
+        const cpmUrl = new URL("https://" + this.cpmURL + "/topic/" + topic);
```
pietrzakacper, line 549: Maybe use util function for doign either http or https request

packages/host/src/lib/cpm-connector.ts
```diff
@@ -180,16 +181,22 @@ export class CPMConnector extends TypedEmitter<Events> {
...
-            server
+            server,
+            https: typeof config.cpmSslCaPath === "string"
+                ? { ca: [fs.readFileSync(config.cpmSslCaPath)] }
+                : undefined
         });
 
         this.logger = new ObjLogger(this);
         this.logger.trace("Initialized.");
+        this.logger.debug("CONNNECTING", config);
```
pietrzakacper, line 199: remove

packages/cli/package.json
```diff
@@ -14,8 +14,7 @@
     "test": "echo no tests yet -- # npm run test:ava",
     "test:ava": "ava",
     "preinstall": "tar --version 2>&1 >/dev/null",
-    "prepack": "node ../../scripts/publish.js",
-    "postbuild": "yarn prepack && cp ./src/completion ../../dist/cli/completion -rL && cp ./src/types/commander-completion.d.ts ../../dist/cli/types/commander-completion.d.ts -L"
```
pietrzakacper, line 18: This will be brought back after it no longer causes CPM to break

packages/host/src/lib/host.ts
```diff
@@ -122,6 +122,7 @@ export class Host implements IComponent {
         const prettyLog = new DataStream().map(prettyPrint({ colors: this.config.logColors }));
 
         this.logger.addOutput(prettyLog);
+        this.serviceDiscovery.logger.addOutput(prettyLog);
```
patuwwy, line 125: this.serviceDiscovery.logger.pipe(this.logger)

packages/types/src/sth-configuration.ts
```diff
@@ -93,6 +93,8 @@ export type STHConfiguration = {
      */
     cpmUrl: string;
 
+    cpmSslCaPath?: string;
```
patuwwy, line 96: missing tsdoc

packages/cli/package.json
patuwwy: no prepack = no /dist/cli

packages/host/src/lib/cpm-connector.ts
```diff
@@ -167,31 +160,64 @@ export class CPMConnector extends TypedEmitter<Events> {
...
-    constructor(cpmHostname: string, cpmId: string, config: CPMConnectorOptions, server: Server) {
+    constructor(private cpmHostname: string, cpmId: string, config: CPMConnectorOptions, server: http.Server) {
         super();
-        this.cpmURL = cpmHostname;
         this.cpmId = cpmId;
         this.config = config;
 
         this.verserClient = new VerserClient({
-            verserUrl: `http://${cpmHostname}/verser`,
+            verserUrl: `${this.cpmUrl}/verser`,
```
alicja-gruzdz, line 181: I see why you proposed a module for managing url :)

packages/host/src/lib/host.ts
```diff
@@ -585,6 +589,7 @@ export class Host implements IComponent {
             }, InstanceMessageCode.INSTANCE_ENDED);
 
             if (csic.provides && csic.provides !== "") {
+                // @TODO not sure why we need that
                 csic.getOutputStream()!.unpipe(this.serviceDiscovery.getData(
```
alicja-gruzdz, line 593: As far as I remember it was introduced following @MichalCz comment that after the Instances termites we should detach its output stream being piped to the topic. I know this is pretty much evident from the code and not a detailed technical explanation but unfortunately I don't remember more. With the new topic design perhaps you will need it more than what we needed it back then, when topic could have been used only once.

alicja-gruzdz (approved):

packages/client-utils/src/types/index.ts
```diff
@@ -43,7 +43,6 @@ export type RequestConfig = {
  * Environmentally independent HttpClient interface.
  */
 export interface HttpClient {
-    fetch?: any;
```
patuwwy, line 46: Please rembember that we are using APIClients (w/ClientUtils) in browser as well.

pietrzakacper: yes, the field remains, it's just initialized in a different way. I think that it doesn't have to be public that's why I removed it from the interface

packages/host/src/lib/host.ts
pietrzakacper: yeah I think it might be to prevent topic closing if the output stream itself has not been ended by the user but only by sequence end. we wil definitely look into that with reusable topics :wink:

packages/cli/package.json
pietrzakacper: Yes, I will bring it back before merging. Actually I wanted to only get rid of postbuild for the time of iterating on the CPM PR

packages/host/src/lib/host.ts
pietrzakacper: good catch

packages/types/src/sth-configuration.ts
pietrzakacper: fixed

packages/host/src/lib/host.ts
alicja-gruzdz: could be, good luck :muscle:

patuwwy (approved):

PR #369, opened 2022-03-24
Add a feature that allows to color logs from stdin

MichalCz:
This allows devs to run a simple command that pretty prints JSON logs (from file or stdin).

Usage:

```
cat logfile | si util log-color
```﻿

We can leave this undocumented (meaning no special info in docs) - there's help in `cli`.

### DIFF
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -1,8 +1,6 @@
 name: Release test
 
-on:
-  pull_request:
-    branches: [ release/*, main ]
+on: workflow_run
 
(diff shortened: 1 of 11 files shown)


jan-warchol (approved):
Awesome!! :star_struck:

jan-warchol:
One thought: maybe name it `parse-log`? Seeing `log-color --no-color` is a bit confusing. :thinking: `parse-log --no-color` would make it obvious.

MichalCz:
Rebased and added to @ErykSol work.

PR #370, opened 2022-03-25
Never close topics

pietrzakacper:
This PR is introducing topics that are always open which effectively makes them reusable.
What we're gaining:
* Topics will always be available for writing/reading, meaning better predictability/reliability of the usage. For example now multiple instances can write to a topic, close the producing stream in any moment and the topic will remain open for further reads/writes even if all data is read.

What we're loosing:
* Now a sequence cannot close the topic, so if some sequence/client behavior depends on its input stream (coming from topic) being ended  these sequences will run indefinitely. For many sequences that might not be a problem. However if needed, "closing" a data stream can be simulated on the data layer. Imagine you want to finish processing, and close some instances that are reading data. You could write a special message to the topic that the reading sequences could react to and close themselves. This has a risk of some unconsumed data remaining in the stream after the end message was sent (imagine one sequence "ends" the topic and the other keeps writing to it). So we might want to revisit this issue and maybe in the future introduce some ways of pausing the topic, or resetting on the topics implementation level.

This is by all means a ***breaking*** change and should be treated accordingly in a release.

### Verify
I added 2 CI tests:
*  ***E2E-013 TC-006 Send data from multiple instances to another instance on the same host***
* ***E2E-013 TC-007 Send and read data two times***

### DIFF
--- a/packages/cli/src/lib/commands/topic.ts
+++ b/packages/cli/src/lib/commands/topic.ts
@@ -14,15 +14,14 @@ export const topic: CommandDefinition = (program) => {
 
     topicCmd.command("send <topic> [<file>]")
         .option("-t, --content-type <value>", "Content-Type", "text/plain")
-        .option("-e, --end", "x-end-stream", false)
         .description("send data to topic")
-        .action(async (topicName, filename, { contentType, end }) => displayEntity(
+        .action(async (topicName, filename, { contentType }) => displayEntity(
             program,
             getHostClient(program).sendNamedData(
                 topicName,
                 filename ? await getReadStreamFromFile(filename) : process.stdin,
                 contentType,
-                end)
+                false)
         ));
 
     topicCmd.command("get <topic>")
--- a/python/reference-apps/python-topic-producer/main.py
+++ b/python/reference-apps/python-topic-producer/main.py
@@ -4,4 +4,4 @@ provides = {
 }
 
 def run(context, input):
-    return input.map(lambda s: f'producer got: {s}').each(print)
+    return input.map(lambda s: f'producer got: {s}' + "\n").each(print)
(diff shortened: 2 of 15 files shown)


python/reference-apps/python-topic-producer/main.py
```diff
@@ -4,4 +4,4 @@
 }
 
 def run(context, input):
-    return input.map(lambda s: f'producer got: {s}').each(print)
+    return input.map(lambda s: f'producer got: {s}' + "\n").each(print)
```
pietrzakacper, line 7: After we no longer close topics, we need this newline character. It's related to a bug in Python runner: https://github.com/scramjetorg/transform-hub/issues/376

patuwwy:
What about sending topic with API?

pietrzakacper:
> What about sending topic with API?

@patuwwy I didn't observe it being closed, but I should make sure in that it's enforced in the implementation and there are tests for that

packages/cli/src/lib/commands/topic.ts
```diff
@@ -14,15 +14,14 @@ export const topic: CommandDefinition = (program) => {
 
     topicCmd.command("send <topic> [<file>]")
         .option("-t, --content-type <value>", "Content-Type", "text/plain")
-        .option("-e, --end", "x-end-stream", false)
         .description("send data to topic")
-        .action(async (topicName, filename, { contentType, end }) => displayEntity(
```
MichalCz, line 19: @pietrzakacper I wouldn't remove this feature and I would actually add another command to explicitly `end` topic, called... uhm... `end`... ;)

It's ok to add it in another PR.

python/reference-apps/python-topic-producer/main.py
MichalCz: Perhaps it would be better to add this to the last line in the actual input stream in runner?

patuwwy (approved):

PR #374, opened 2022-03-28
Add config endpoint

jan-warchol:
For debugging purposes it's worth to see the STH/Host config. This endpoint shows the safe to show values.

```
/api/v1/config
```

Verify:

```
yarn && yarn build
yarn start
```

And then in another terminal:

```
curl http://localhost:8000/api/v1/config
```

Config should be missing `authConfigPath` in Kubernetes and `cpmSslCaPath` on config level and sequencesRoot on both levels.

### DIFF
--- a/packages/sth-config/src/config-service.ts
+++ b/packages/sth-config/src/config-service.ts
@@ -1,4 +1,4 @@
-import { DeepPartial, STHConfiguration } from "@scramjet/types";
+import { DeepPartial, PublicSTHConfiguration, STHConfiguration } from "@scramjet/types";
 import { merge } from "@scramjet/utility";
 import path from "path";
 import { homedir } from "os";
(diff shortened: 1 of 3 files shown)


Tatarinho (approved):

packages/sth-config/src/config-service.ts
```diff
@@ -83,4 +83,17 @@ export class ConfigService {
     update(config: DeepPartial<STHConfiguration>) {
         merge(this.config, config);
     }
+
+    static getConfigInfo(config: STHConfiguration): PublicSTHConfiguration {
+        const {
+            kubernetes: kubeFull,
+            sequencesRoot: optionsSequencesRoot2,
```
jan-warchol, line 90: shouldn't this be `optionsSequencesRoot`?

jan-warchol: [According to Michał](https://scramjetcloudplatform.slack.com/archives/C02FC09DNUV/p1648478075077029?thread_ts=1648470345.103749&cid=C02FC09DNUV) there are two independent dummy variables, so this is ok

PR #386, opened 2022-03-30
Build MW ApiClient - iife/cjs/esm

patuwwy:


kociolekscramjet (approved):
Works as intended :)

patuwwy:
@kociolekscramjet how do you use it

kociolekscramjet:
need some more time for research

PR #392, opened 2022-04-01
Fix for adding Python sequence library to path

Tatarinho:


MichalCz (approved):

packages/adapters/src/process-instance-adapter.ts
```diff
@@ -83,7 +83,7 @@ IComponent {
             path.resolve(__dirname, require.resolve("@scramjet/python-runner"), "..")
         );
 
-        pythonpath += `:${process.env.PYTHONPATH}:${sequenceDir}`;
+        pythonpath += `:${process.env.PYTHONPATH}:${sequenceDir}/dist/__pypackages__`;
```
MichalCz, line 86:
```suggestion
        pythonpath += `:${process.env.PYTHONPATH}:${sequenceDir}/__pypackages__`;
```

MichalCz (changes requested):
This needs fixing since in the example solution the package is not packed as it should be.

packages/adapters/src/process-instance-adapter.ts
Tatarinho: Same change for Dockerfile I guess

MichalCz (approved):

PR #398, opened 2022-04-05
Introduce URLs normalization

### LINKED ISSUE #372: Inconsistent/confusing protocol handling in api urls
**Describe the bug**

When working with transform hub (and cloud platform), sometimes I have to include `http://` in the urls, and sometimes I must _not_ include it. It's confusing.

**Example: CPM url**
`--cpm-url` option for STH doesn't work when the url already starts with `http://` - it seems that the protocol is always prepended. The result is that STH cannot connect.
**Expected behavior**: either prepend protocol to the url only when it's not specified, or throw an error when the user provides url with protocol.

**Example: HostClient url**
`HostClient` does not work when the provided url doesn't have the protocol - code like this:
```
import { HostClient } from "@scramjet/api-client";
const client = new HostClient("localhost:8000/api/v1")
```
will result in a difficult to understand error like to this:
```
Error: Failed to request 'localhost:8000/api/v1/load-check' with code undefined.
          at Function.from (/home/jan/src/platform-tests/node_modules/@scramjet/src/client-error.ts:77:20)
          at ClientUtils.safeRequest (/home/jan/src/platform-tests/node_modules/@scramjet/src/client-utils.ts:96:31)
          at async default_1.<anonymous> (/home/jan/src/platform-tests/steps/steps.ts:13:13)
```
On the other hand, when you omit protocol in CLI, you get a more helpful (and immediate) error:
```
$ si config apiUrl localhost:8000/api/v1
$ si seq list
Request failed with code "undefined" status: undefined
Error: Only HTTP(S) protocols are supported
    at Function.from (/home/jan/src/transform-hub/dist/cli/node_modules/@scramjet/client-utils/client-error.js:69:16)
    at /home/jan/src/transform-hub/dist/cli/node_modules/@scramjet/client-utils/client-utils.js:36:46
    at async displayEntity (/home/jan/src/transform-hub/dist/cli/lib/output.js:30:17)
```
(only the "`undefined`" part is a bit confusing).

**Expected behavior**: either prepend protocol to the url when it's not specified, or throw an error immediately when instantiating `HostClient` with wrong url.

In general, I would expect that both places behaved similarly - either the protocol is required or not.

### LINKED ISSUE #373: API url with trailing slash is accepted but doesn't work
**Describe the bug**
API url with trailing slash does not work, but neither `si` nor `HostClient` complain when it's specified.

**Steps to Reproduce**
```js
import { HostClient } from "@scramjet/api-client";
const client = new HostClient("localhost:8000/api/v1/")  // no error
client.getLoadCheck()  // -> this will try to request "localhost:8000/api/v1//load-check"
```
or
```bash
$ si config apiUrl http://localhost:8000/api/v1/
$ si seq list
Request failed with code "404" status: undefined
Error: Failed to request 'http://localhost:8000/api/v1//sequences' with code 404.
# [...]
```

**Expected behavior**
Either handle duplicated slashes (github does that - `https://github.com///scramjetorg//transform-hub` works just like `https://github.com/scramjetorg/transform-hub/`) or throw an error when an api url with a trailing slash is provided.

f1ames:
This PR introduces URLs normalization mechanism in `ClientUtils` package (used by all `*Client`) and also in Host for `cpmUrl` config option.

Internally it uses [`normalize-url` library](https://www.npmjs.com/package/normalize-url).

## How to test?

**For `HostClient`:**

```bash
yarn start -P 8000
```

```js
import { HostClient } from "./dist/api-client/index.js";

const client1 = new HostClient("localhost:8000/api/v1");
client1.getLoadCheck().then(console.log);

const client2 = new HostClient("localhost:8000/api/v1/");
client2.getLoadCheck().then(console.log);

const client3 = new HostClient("http://localhost:8000/api/v1/");
client3.getLoadCheck().then(console.log);

const client4 = new HostClient("http://localhost:8000/api/v1");
client4.getLoadCheck().then(console.log);

const client5 = new HostClient("localhost:8000/api/v1//");
client5.getLoadCheck().then(console.log);
```

All request should return `200` response with Host load data. It defaults to `http:` protocol. Haven't checked `https:` since I don't have it set up for now.

**For `si`:**

```bash
yarn install && yarn build:all && npm i -g ./dist/cli
```

```js
si config apiUrl http://localhost:8000/api/v1/
si seq list

si config apiUrl localhost:8000/api/v1
si seq list

si config apiUrl http://localhost:8000//api//v1
si seq list
```
All of the above should work, returning sequences list (empty array `[]` if nothing was uploaded).

**For `Host` `cpmUrl` option**:

```
yarn start -P 8000 --cpm-url http://localhost:9000
yarn start -P 8000 --cpm-url localhost:9000
```

---

Closes #372.
Closes #373.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -25,6 +25,7 @@
     "@scramjet/client-utils": "^0.19.2",
     "@scramjet/middleware-api-client": "^0.19.2",
     "@scramjet/obj-logger": "^0.19.0",
+    "@scramjet/utility": "^0.19.2",
     "commander": "^8.3.0",
     "commander-completion": "^1.0.1",
     "find-package-json": "^1.2.0",
--- a/packages/cli/src/lib/config.ts
+++ b/packages/cli/src/lib/config.ts
@@ -1,4 +1,5 @@
 import { existsSync, writeFileSync, readFileSync } from "fs";
+import { normalizeUrl } from "@scramjet/utility";
 import { globalConfigFile, sessionConfigFile } from "./paths";
 import { configEnv, configFormat, GlobalConfigEntity, isConfigEnv, isConfigFormat, SessionConfigEntity } from "../types";
 
@@ -161,7 +162,7 @@ class GlobalConfig extends DefaultFileConfig {
     }
 
     setApiUrl(apiUrl: string): boolean {
-        return this.setConfigValue("apiUrl", apiUrl) as boolean;
+        return this.setConfigValue("apiUrl", normalizeUrl(apiUrl)) as boolean;
     }
     setDebug(debug: boolean): boolean {
         return this.setConfigValue("debug", debug) as boolean;
@@ -170,7 +171,7 @@ class GlobalConfig extends DefaultFileConfig {
         return this.setConfigValue("format", format) as boolean;
     }
     setMiddlewareApiUrl(middlewareApiUrl: string): boolean {
-        return this.setConfigValue("middlewareApiUrl", middlewareApiUrl) as boolean;
+        return this.setConfigValue("middlewareApiUrl", normalizeUrl(middlewareApiUrl)) as boolean;
     }
     setEnv(env: configEnv): boolean {
         return this.setConfigValue("env", env) as boolean;
--- a/packages/client-utils/package.json
+++ b/packages/client-utils/package.json
@@ -24,7 +24,7 @@
     "n-readlines": "^1.0.1",
     "node-fetch": "^2.6.7",
     "scramjet": "^4.36.6",
-    "normalize-url": "^7.0.3"
+    "normalize-url": "^5.3.1"
   },
   "devDependencies": {
     "@scramjet/types": "^0.19.2",
--- a/packages/client-utils/package.json
+++ b/packages/client-utils/package.json
@@ -20,6 +20,7 @@
     "@scramjet/obj-logger": "^0.19.2",
     "@scramjet/sth-config": "^0.19.2",
     "@scramjet/symbols": "^0.19.2",
+    "@scramjet/utility": "^0.19.2",
     "n-readlines": "^1.0.1",
     "node-fetch": "^2.6.7",
     "scramjet": "^4.36.6"
--- a/packages/client-utils/package.json
+++ b/packages/client-utils/package.json
@@ -23,7 +23,8 @@
     "@scramjet/utility": "^0.19.2",
     "n-readlines": "^1.0.1",
     "node-fetch": "^2.6.7",
-    "scramjet": "^4.36.6"
+    "scramjet": "^4.36.6",
+    "normalize-url": "^7.0.3"
   },
   "devDependencies": {
     "@scramjet/types": "^0.19.2",
--- a/packages/client-utils/src/client-utils.ts
+++ b/packages/client-utils/src/client-utils.ts
@@ -1,20 +1,22 @@
 /* eslint-disable no-console */
 import { ClientError, QueryError } from "./client-error";
 import { Headers, HttpClient, RequestLogger, SendStreamOptions, RequestConfig } from "./types";
-import { normalizeUrl } from "@scramjet/utility";
 
 /**
  * Provides HTTP communication methods.
  */
 export abstract class ClientUtilsBase implements HttpClient {
     private log?: RequestLogger;
+    private normalizeUrlFn: (url: string) => string;
 
     static headers: Headers = {};
 
     constructor(
         public apiBase: string,
-        private fetch: any
+        private fetch: any,
+        normalizeUrlFn?: (url: string) => string
     ) {
+        this.normalizeUrlFn = normalizeUrlFn || ((url: string) => url);
     }
 
     /**
@@ -105,7 +107,7 @@ export abstract class ClientUtilsBase implements HttpClient {
      * @returns {Promise<T>} Promise resolving to given type.
      */
     async get<T>(url: string): Promise<T> {
-        return this.safeRequest<T>(normalizeUrl(`${this.apiBase}/${url}`), {}, { parse: "json" });
+        return this.safeRequest<T>(this.normalizeUrlFn(`${this.apiBase}/${url}`), {}, { parse: "json" });
     }
 
     /**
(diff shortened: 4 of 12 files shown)


patuwwy (approved):

Tatarinho (approved):

alicja-gruzdz (approved):
:heavy_check_mark: CLI
Zrzut ekranu z 2022-04-07 09-52-11

jan-warchol (commented):
I've noticed two things that may be worth addressing, but I guess they aren't blockers:

1) normalized url is used for making request, but `si` config keeps whatever was provided verbatim:
```
$ si config apiUrl localhost:8000/api/v1///
$ si seq list
Request ok: http://localhost:8000/api/v1/sequences status: 200 OK
$ si config get apiUrl
'localhost:8000/api/v1///'
```

2) `apiBase` attribute of the client is not normalized (should it?), and - more interestingly - it differs between `HostClient` and its `ClientUtils`:

```
> import { HostClient } from "./dist/api-client/index.js";
> const client = new HostClient("localhost:8000/api/v1///");
> client
HostClient {
  apiBase: 'localhost:8000/api/v1//',
  client: ClientUtils {
    apiBase: 'localhost:8000/api/v1///',
    fetch: [Function (anonymous)],
    normalizeUrlFn: [Function: normalizeUrl]
  }
}
```

jan-warchol (commented):
I've noticed two things that may be worth addressing, but I guess they aren't blockers:

1) normalized url is used for making request, but `si` config keeps whatever was provided verbatim:
```
$ si config apiUrl localhost:8000/api/v1///
$ si seq list
Request ok: http://localhost:8000/api/v1/sequences status: 200 OK
$ si config get apiUrl
'localhost:8000/api/v1///'
```

2) `apiBase` attribute of the client is not normalized (should it?), and - more interestingly - it differs between `HostClient` and its `ClientUtils`:

```
> import { HostClient } from "./dist/api-client/index.js";
> const client = new HostClient("localhost:8000/api/v1///");
> client
HostClient {
  apiBase: 'localhost:8000/api/v1//',
  client: ClientUtils {
    apiBase: 'localhost:8000/api/v1///',
    fetch: [Function (anonymous)],
    normalizeUrlFn: [Function: normalizeUrl]
  }
}
```

jan-warchol (commented):
I've noticed two things that may be worth addressing, but I guess they aren't blockers:

1) normalized url is used for making request, but `si` config keeps whatever was provided verbatim:
```
$ si config apiUrl localhost:8000/api/v1///
$ si seq list
Request ok: http://localhost:8000/api/v1/sequences status: 200 OK
$ si config get apiUrl
'localhost:8000/api/v1///'
```

2) `apiBase` attribute of the client is not normalized (should it?), and - more interestingly - it differs between `HostClient` and its `ClientUtils`:

```
> import { HostClient } from "./dist/api-client/index.js";
> const client = new HostClient("localhost:8000/api/v1///");
> client
HostClient {
  apiBase: 'localhost:8000/api/v1//',
  client: ClientUtils {
    apiBase: 'localhost:8000/api/v1///',
    fetch: [Function (anonymous)],
    normalizeUrlFn: [Function: normalizeUrl]
  }
}
```

jan-warchol:
Another nice-to-have: if you provide invalid url to `si config`, it is accepted, but then there's a failure when you try to use it. It would be great if it complained immediately.

```
$ si config apiUrl :8000/api/v1///

$ si seq list
TypeError [ERR_INVALID_URL]: Invalid URL
    at new NodeError (node:internal/errors:371:5)
    at onParseError (node:internal/url:552:9)
    at new URL (node:internal/url:628:5)
    at normalizeUrl (/home/jan/src/transform-hub/dist/utility/node_modules/normalize-url/index.js:98:17)
    at ClientUtils.normalizeUrl [as normalizeUrlFn] (/home/jan/src/transform-hub/dist/utility/normalize-url.js:10:40)
    at ClientUtils.get (/home/jan/src/transform-hub/dist/client-utils/client-utils.js:64:38)
    at HostClient.listSequences (/home/jan/src/transform-hub/dist/api-client/host-client.js:12:28)
    at Command.<anonymous> (/home/jan/src/transform-hub/dist/cli/lib/commands/sequence.js:51:103)
    at Command.listener [as _actionHandler] (/home/jan/src/transform-hub/dist/cli/node_modules/commander/lib/command.js:488:17)
    at /home/jan/src/transform-hub/dist/cli/node_modules/commander/lib/command.js:1227:65 {
  input: 'http::8000/api/v1////sequences',
  code: 'ERR_INVALID_URL'
}
```

f1ames:
> 1. normalized url is used for making request, but `si` config keeps whatever was provided verbatim:

Yes, it's normalized in `ClientUtils` when sending a request. I guess it would make sense to normalize it also for storing in config. Can't think of a case where it could broke something really :thinking: **I will fix that :+1:** 

> 2\. `apiBase` attribute of the client is not normalized (should it?),

Same as above, it's normalized in `ClientUtils` when sending a request. I assumed it doesn't make sense to normalize same URL (or it's parts) in multiple places. Maybe at some point one may try to use `HostClient.apiBase` so better to store it in normalized form too :thinking: WDYT @jan-warchol?

> and - more interestingly - it differs between `HostClient` and its `ClientUtils`

Hmm, interesting indeed. I would say on of those do some initial manipulation with this `apiBase` maybe.

jan-warchol:
> > 1. normalized url is used for making request, but `si` config keeps whatever was provided verbatim:
> 
> Yes, it's normalized in `ClientUtils` when sending a request. I guess it would make sense to normalize it also for storing in config. Can't think of a case where it could broke something really thinking **I will fix that +1**

awesome :+1: 

> > 2. `apiBase` attribute of the client is not normalized (should it?),
> 
> Same as above, it's normalized in `ClientUtils` when sending a request. I assumed it doesn't make sense to normalize same URL (or it's parts) in multiple places. Maybe at some point one may try to use `HostClient.apiBase` so better to store it in normalized form too thinking WDYT @jan-warchol?

Hmm. Probably. But I also think it's quite minor and probably not worth the time. :wink:

f1ames:
> > > 1. normalized url is used for making request, but `si` config keeps whatever was provided verbatim:
> > 
> > 
> > Yes, it's normalized in `ClientUtils` when sending a request. I guess it would make sense to normalize it also for storing in config. Can't think of a case where it could broke something really thinking **I will fix that +1**
> 
> awesome +1

@jan-warchol, please see https://github.com/scramjetorg/transform-hub/pull/398/commits/10c988a74368731aebe7b7eb5e58490b7a1bd650.

PR #400, opened 2022-04-05
Handle Runner init errors in CSIController and Clients.

pietrzakacper:
Now when Runner fails on start because for example the sequence path is incorrect, it will exit with the appropriate exit code, then this exit code will be mapped in csi controller to error message and sent out in REST API response to sequence start request

### Verify
Make sure you have the CLI installed from this branch and STH running.
1. Modify package.json of [some sample](https://github.com/scramjetorg/scramjet-cloud-docs/tree/main/samples/hello) to an invalid one
2. Pack, send and start the sequence
```
si pack . -o dist.tar.gz && si seq send dist.tar.gz && si seq start -
```
3. You should see an error in CLI similar to this:
image

### DIFF
--- a/packages/runner/src/bin/start-runner.ts
+++ b/packages/runner/src/bin/start-runner.ts
@@ -4,6 +4,7 @@ import { Runner } from "../runner";
 import fs from "fs";
 import { AppConfig } from "@scramjet/types";
 import { HostClient } from "../host-client";
+import { RunnerExitCode } from "@scramjet/symbols";
 
 const sequencePath: string = process.env.SEQUENCE_PATH?.replace(/.js$/, "") + ".js";
 const instancesServerPort = process.env.INSTANCES_SERVER_PORT;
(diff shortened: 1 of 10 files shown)


packages/runner/src/bin/start-runner.ts
```diff
@@ -12,22 +13,22 @@ const instanceId = process.env.INSTANCE_ID;
 
 if (!instancesServerPort || instancesServerPort !== parseInt(instancesServerPort, 10).toString()) {
     console.error("Incorrect run argument: instancesServerPort");
-    process.exit(1);
+    process.exit(RunnerExitCode.INVALID_ENV_VARS);
```
MichalCz, line 16: @pietrzakacper providing exit code this way is deprecated, it should be done via `process.exitCode = `

pietrzakacper: I dont think it's deprecated. Setting a property is just a way for setting exitCode when we don't want to exit forcefully. https://nodejs.org/api/process.html#processexitcode

pietrzakacper: In this case, we want to

MichalCz: Oki. I was mistaken then. :)

f1ames (approved):
LGTM :+1: 

---

Tested with:

```
yarn start --runtime-adapter process
```

and then sending sequence with invalid `main` field in `package.json`.

Via `si`:


image


Via API:


image

---

@pietrzakacper from what I understand it's only about invalid `main` field? I tried with differently broken sequences (empty `index.js`, syntax error, invalid tar) but the errors then are in different places (before runner initialization or after runner is started) so it's not in the scope of this PR right?

MichalCz (approved):

PR #402, opened 2022-04-06
Improve version endpoint to better identify running service

f1ames:
This PR improves `/version` endpoint response for host.

Previously it returned only `version` pointing to `package.json` version. Now it returns additional information to help better identify the service underneath:

```js
{
    "service": "host",   // service name
    "version": "0.19.2", // service version
    "apiVersion": "v1",  // service API version
    "build": "87fc6350"  // service build - git hash from which service is run
}
```

Git hash (`build`) is updated as a prebuild step in host package.

## How to test?

1. Start host:

```bash
yarn start     # build version
yarn start:dev # dev version
```

2. Query `/version` endpoint, for example:

```bash
curl localhost:8000/api/v1/version
```

It should return something like:

```
{"service":"host","apiVersion":"v1","version":"0.19.2","build":"87fc6350"}
```

## Naming

I'm not really convinced about naming, especially `host` (this is our official naming but may sound generic for someone not familiar with it) and `version` vs `apiVersion`. So any suggestions welcome.

---

If this is the direction we would like to go, I will update MW and potentially Manager, MM and MH in a similar way.

### DIFF
--- a/scripts/hash-to-file.js
+++ b/scripts/hash-to-file.js
@@ -8,10 +8,14 @@ const { join, resolve } = require("path");
 const exec = promisify(execCallback);
 
 (async () => {
-    const toPath = process.argv[2] || "";
+    try {
+        const toPath = process.argv[2] || "";
 
-    const hash = await exec("git rev-parse --short HEAD");
-    const contents = `export const hash = "${hash.stdout.trim()}";\n`;
+        const hash = await exec("git rev-parse --short HEAD");
+        const contents = `export const hash = "${hash.stdout.trim()}";\n`;
 
-    await writeFile(resolve(join("./", toPath, "hash.ts")), contents, "utf8");
+        await writeFile(resolve(join("./", toPath, "hash.ts")), contents, "utf8");
+    } catch (err) {
+        console.log(`Saving git hash to file failed: ${err}`);
+    }
 })();
--- a/scripts/hash-to-file.js
+++ b/scripts/hash-to-file.js
@@ -12,9 +12,9 @@ const exec = promisify(execCallback);
         const toPath = process.argv[2] || "";
 
         const hash = await exec("git rev-parse --short HEAD");
-        const contents = `export const hash = "${hash.stdout.trim()}";\n`;
+        const contents = { hash: hash.stdout.trim(), timestamp: Date.now() };
 
-        await writeFile(resolve(join("./", toPath, "hash.ts")), contents, "utf8");
+        await writeFile(resolve(join("./", toPath, "build.info.json")), JSON.stringify(contents, null, "  "), "utf8");
     } catch (err) {
         console.log(`Saving git hash to file failed: ${err}`);
     }
--- a/scripts/hash-to-file.js
+++ b/scripts/hash-to-file.js
@@ -1,0 +1,17 @@
+#!/usr/bin/env node
+/* eslint-disable @typescript-eslint/no-floating-promises */
+const { writeFile } = require("fs/promises");
+const { exec: execCallback } = require("child_process");
+const { promisify } = require("util");
+const { join, resolve } = require("path");
+
+const exec = promisify(execCallback);
+
+(async () => {
+    const toPath = process.argv[2] || "";
+
+    const hash = await exec("git rev-parse --short HEAD");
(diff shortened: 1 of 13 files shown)


patuwwy (approved):

alicja-gruzdz:
> 'm not really convinced about naming, especially `host` (this is our official naming but may sound generic for someone not familiar with it) and `version` vs `apiVersion`. So any suggestions welcome.

About the `host`: one alternative could be to add to `package.json` a proper name like `Scramjet Tranform Hub` and read it from there... or take  `"name": "@scramjet/host"` from the `package.json`

jan-warchol:
> > 'm not really convinced about naming, especially `host` (this is our official naming but may sound generic for someone not familiar with it) and `version` vs `apiVersion`. So any suggestions welcome.
> 
> About the `host`: one alternative could be to add to `package.json` a proper name like `Scramjet Tranform Hub` and read it from there... or take `"name": "@scramjet/host"` from the `package.json`

I think `Scramjet Tranform Hub` would be most meaningful for users, but also slightly misleading, as sth != host. As much as I would love to make some changes in the naming, I think the best choice now is `@scramjet/host`.

As for `version` and `apiVersion`, I think they are fine.

scripts/hash-to-file.js
```diff
@@ -0,0 +1,21 @@
...
+const exec = promisify(execCallback);
+
+(async () => {
+    try {
+        const toPath = process.argv[2] || "";
+
+        const hash = await exec("git rev-parse --short HEAD");
+        const contents = `export const hash = "${hash.stdout.trim()}";\n`;
+
+        await writeFile(resolve(join("./", toPath, "hash.ts")), contents, "utf8");
```
jan-warchol, line 17: I see one problem: this updates a file that is tracked by git, so running a build results in "modified" repository state :/
I think that instead we should put this file together with other generated files in `dist`. Perhaps it shouldn't be importable, but a json or something.

jan-warchol (changes requested):
Unfortunately I've found one problem :/

scripts/hash-to-file.js
f1ames: Yes, I was thinking about it. Tried to ignore it at first but since it's imported it needs to be there (if not, running via `ts-node` will fail). Also thought about putting it into`dist/` but since the paths may be relative different, e.g. when `sth` is used as submodule, it may not be reliable.

How about putting it as ignored json file in package `src` and then having kind of util which will read from it during runtime? And it will be simple copied to dist during build?

jan-warchol: What about not importing it, but just reading its contents (only if it's present)?
The thing with ignored file in `src` doesn't seem right to me... and it won't be there before the first build, so you need to handle the case when the file is absent anyway :D

f1ames: I changed it in the meantime trying to make it similar to what you had described - so having ignored json file with version hash which is the read during runtime. It also handles case where the file is absent. Please take a look at - https://github.com/scramjetorg/transform-hub/pull/402/commits/eba5f3276ab1d39589eb6101af5df0dd153cb0db.

jan-warchol: Yeah, that's more or less what I meant.
It seems that if someone builds one commit, and then runs from source from another commit, `build` will still report the first commit ID. That would be quite confusing. I suggest not to keep this file with sources, but put it directly to `dist`.

f1ames: > It seems that if someone builds one commit, and then runs from source from another commit, `build` will still report the first commit ID. That would be quite confusing. I suggest not to keep this file with sources, but put it directly to `dist`.

Ok, I get the point here. So for source it could return `unknown` hash. Actually, I changed it to `source` which should be meaningful in this case. So now `build.info.json` is generated, moved to dist and removed from `src` during build step. I also left `savehash` script intentionally (which only generates the file) so one can test/run it - it is also required for one CI job to be run before build, since build itself is done on Docker container with no `git` available. See https://github.com/scramjetorg/transform-hub/pull/402/commits/be8f980665b64985ace4fcaf4abfeca6355db84a.

jan-warchol (approved):

scripts/hash-to-file.js
jan-warchol: Thanks!

PR #405, opened 2022-04-06
Image choice for Kubernates adapter

Tatarinho:
As part of PR:
1. Support for image choice in Kubernates adapter
2. As default, we use **scramjetorg/runner** and **scramjetorg/runner-py** for creating pods
3. Python-Runner adjustments for Kubernates
4. New CLI k8s parameters:
```
--k8s-runner-image=repo.example.com/runner:0.19.2
--k8s-runner-py-image=repo.example.com/runner-py:0.19.2
```
5. @daro1337 asked for parameter "timeout", which won't kill pod immediately after finishing sequence/instance. Useful for developing. I'm not only sure if the name for that is not slightly confusing, usage: 
`--k8s-timeout=5000`

**Update**
timeout parameter name changed according to @daro1337 suggestion
`--k8s-runner-cleanup-timeout=5000`

### DIFF
--- a/packages/adapters/src/kubernetes-instance-adapter.ts
+++ b/packages/adapters/src/kubernetes-instance-adapter.ts
@@ -84,9 +84,9 @@ IComponent {
             INSTANCE_ID: instanceId,
         }).map(([name, value]) => ({ name, value }));
 
-        const runnerImage = config.engines.node
-            ? this.adapterConfig.runnerImages.node
-            : this.adapterConfig.runnerImages.python3;
+        const runnerImage = config.engines.python3
+            ? this.adapterConfig.runnerImages.python3
+            : this.adapterConfig.runnerImages.node;
 
         await this.kubeClient.createPod(
             {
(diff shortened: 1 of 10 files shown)


daro1337:
> 5\. --k8s-timeout=5000

Well, maybe `—k8s-runner-cleanup-timeout`?
Thanks, for this feature!

Tatarinho:
> > 5. --k8s-timeout=5000
> 
> Well, maybe `—k8s-runner-cleanup-timeout`? Thanks, for this feature!

done

daro1337 (approved):

packages/adapters/src/kubernetes-instance-adapter.ts
```diff
@@ -84,6 +84,10 @@ IComponent {
             INSTANCE_ID: instanceId,
         }).map(([name, value]) => ({ name, value }));
 
+        const runnerImage = config.engines.node
```
pietrzakacper, line 87: Other instance adapters actually have a reversed logic. They switch to python3 if it's present in the engines config. I think this behavior should resemble that to avoid situation in which sequence without an engine specified will default to Python (I guess we want Node to be the default)

pietrzakacper (changes requested):

pietrzakacper (approved):

PR #420, opened 2022-04-11
docs: How to write a sequence?

pietrzakacper:
## TODO
- [ ] Test parameters
- [ ] Test all sequence example code
- [ ] Decide whether to inform about topics scope
- [ ] Link CLI, API and Clients docs
- [ ] Decide what else should be described? (events, kill handler etc.)


docs/read-more/how-to-write-a-sequence.md
```diff
@@ -0,0 +1,221 @@
...
+## Producing data (output stream)
+To stream data from a sequence we need it to return values over time. Some constructs in JavaScript that enable that are NodeJS streams, Generators and Iterables. Whatever you return from your sequence will be your ***output*** stream. We can choose whatever the solution is right for us. Here are some examples of a sequence producing a stream of integers every second.  
+Using a stream:
+```ts
+export default function() {
+    const out = new PassThrough()
+
+    let i = 0
+    setInterval(() => {
+        out.write(i++)
```
MichalCz, line 34: We need to add drain logic here. Otherwise our customers will write programs that exhaust all the memory rather quickly

docs/read-more/how-to-write-a-sequence.md
```diff
@@ -0,0 +1,221 @@
...
+
+```ts
+/** The function parameters are
+* input stream
+* ...params passed to instance on start
+*/
+export default function(input, param1, param2) {
+    const out = new PassThrough()
+
+    input.on('data', (data) => {
```
MichalCz, line 12: Let's put asnyc iteration version higher. Also we need flow control in this sample

docs/read-more/how-to-write-a-sequence.md
```diff
@@ -0,0 +1,221 @@
...
+Transforming data is really a combination of consuming and producing, usually with some logic in between. Let’s filter the incoming input stream of numbers to include only the even ones. 
+
+Using streams:
+```ts
+export default function(input) {
+    const out = new PassThrough()
+
+    input.on('data', (num) => {
+        if(num % 2 === 0) {
+            out.write(num)
```
MichalCz, line 101: Again, and downwards we need flow control

MichalCz:
@a-tylenda @agne6ka we should get this on our website as soon as Kacper is done with this.

docs/read-more/how-to-write-a-sequence.md
pietrzakacper: fixed

docs/read-more/how-to-write-a-sequence.md
pietrzakacper: fixed

docs/read-more/how-to-write-a-sequence.md
pietrzakacper: fixed

docs/read-more/how-to-write-a-sequence.md
```diff
@@ -0,0 +1,251 @@
...
+    { requires: 'hello', contentType: 'text/plain' },
+    function(input) { ... }
+]
+```
+
+### Interacting with topics
+Apart from sequences communicating between each other you can also feed/consume a topic using our CLI, REST API, or API Client. 
+
+## Standard streams (stdin/stdout/stderr)
+Every sequence has access to standard streams of a program. You can read data from stdin. Send additional information to stdout and to stderr. These are separate from input/output streams.
```
alicja-gruzdz, line 218: Perhaps it's obvious for some but I'd add that e.g. console.log(..) goes to /stdout because I've seen this confusion especially when /log enters the game :)

docs/read-more/how-to-write-a-sequence.md
```diff
@@ -0,0 +1,251 @@
...
+}
+```
+### Reading from a topic 
+To send topic as input stream we specify a topic config object with two properites: `requires` - with a name of topic, and `contentType` for ensuring proper encoding.
+```ts
+const CRITICAL_TEMP_CELCIUS = 40
+
+const app = [
+    {
+        requires: 'temperature-readouts',
```
alicja-gruzdz, line 190: You described topics very well.
I have a question to @MichalCz about our support for defining topics within the Sequence code, because he mentioned on our last Friday's meeting we should deprecate it and only support defining topics when starting an Instance the way you have implemented it recently. So the question is whether we want to teach users to write the Sequences that way.

alicja-gruzdz (approved):
I really enjoyed reading it. It's a clear and user friendly content.

docs/read-more/how-to-write-a-sequence.md
```diff
@@ -0,0 +1,251 @@
...
+    return out
+}
+```
+
+### Typescript
+Sequences that only produce data should be typed as [ReadableApp](https://hub.scramjet.org/docs/types/modules#readableapp).  
+[Here's an example](https://github.com/scramjetorg/scramjet-cloud-docs/blob/main/samples/scraping/app.ts).
+
+### Reading from output stream
+You could read this stream using our CLI, REST API, or API Client.
```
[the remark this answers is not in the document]
pietrzakacper: fixed

docs/read-more/how-to-write-a-sequence.md
```diff
@@ -0,0 +1,251 @@
...
+    const fn = () => {
+        const canWrite = out.write(i++)
+        if(!canWrite) {
+            clearInterval(intervalRef)
+        }
+    }
+
+    let intervalRef = setInterval(fn, 1000)
+
+    out.on('drain', () => {
```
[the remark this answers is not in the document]
pietrzakacper: fixed

docs/read-more/how-to-write-a-sequence.md
```diff
@@ -0,0 +1,251 @@
...
+}
+```
+
+### Input stream encoding
+Note that input stream is special in a way it already encodes your stream using a specified contentType.
+In this case if you wanted to send the stream using our CLI it would look like this:
+```bash
+si inst input <instance-id> --content-type "application/x-ndjson"
+```
+Then you can send it JSON strings separated by newline characters.
```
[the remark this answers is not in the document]
pietrzakacper: fixed

docs/read-more/how-to-write-a-sequence.md
```diff
@@ -0,0 +1,251 @@
...
+        this.logger.debug('Input stream ended')
+    })
+
+    // it returns an output stream
+    return out
+}
+```
+
+## Producing data (output stream)
+To stream data from a sequence we need it to return values over time. Some constructs in JavaScript that enable that are NodeJS streams, Generators and Iterables. Whatever you return from your sequence will be your ***output*** stream. We can choose whatever the solution is right for us.
```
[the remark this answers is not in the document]
pietrzakacper: fixed

docs/read-more/how-to-write-a-sequence.md
pietrzakacper: fixed

docs/read-more/how-to-write-a-sequence.md
pietrzakacper: Will ask @MichalCz about that

alicja-gruzdz: thanks

PR #425, opened 2022-04-12
Introduce: si seq prune

MichalCz:
There's not much to it. Upload a couple sequences and then:

```sh
si seq ls
si seq prune
si seq ls
```

The second `ls` should show... no sequences? :)

### DIFF
--- a/packages/cli/src/lib/commands/sequence.ts
+++ b/packages/cli/src/lib/commands/sequence.ts
@@ -3,9 +3,9 @@ import { lstatSync } from "fs";
 import { readFile } from "fs/promises";
 import { CommandDefinition } from "../../types";
 import { isDevelopment } from "../../utils/isDevelopment";
-import { getHostClient, getReadStreamFromFile, packAction } from "../common";
+import { getHostClient, getInstance, getReadStreamFromFile, packAction } from "../common";
 import { getPackagePath, getSequenceId, sessionConfig } from "../config";
-import { displayEntity, displayObject } from "../output";
+import { displayEntity, displayMessage, displayObject } from "../output";
 
 const sendPackage = async (sequencePackage: string) => {
     const seq = await getHostClient().sendSequence(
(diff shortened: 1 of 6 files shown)


ErykSol (changes requested):

packages/cli/src/lib/commands/sequence.ts
```diff
@@ -69,6 +69,30 @@ export const sequence: CommandDefinition = (program) => {
         .description("create archived file (package) with sequence for later use")
         .action((path, { stdout, output }) => packAction(path, { stdout, output }));
 
+    sequenceCmd
```
ErykSol, line 72: At the end of this file is already prepared command for prune in developer mode- this should be moved down and merged with that one (remove developer flag)  to keep proper command order.

ErykSol (changes requested):

ErykSol (approved):

packages/cli/src/lib/commands/sequence.ts
```diff
@@ -69,6 +69,40 @@ export const sequence: CommandDefinition = (program) => {
...
+        .command("prune")
```
a-tylenda:
```suggestion
        .command("prune")
        .description("delete multiple sequences on actual selected hub")
```

a-tylenda:
Verification failed during command execution: `si seq prune --force`.
**Steps to Reproduce**
1. start sth: `scramjet-transform-hub`
2. send a few Sequences to STH, start the last one: `si seq start -`
3. force prune to delete all Sequences including the one which is running: `si seq prune --force`

Screenshot of error I received:

image

I think that there is not enough time for the Sequence to stop working after "kill" is sent. DELETE request is sent while killing process is not finished.

packages/cli/src/lib/commands/sequence.ts
```diff
@@ -140,15 +140,37 @@ export const sequence: CommandDefinition = (program) => {
...
+                        displayMessage(`Sequence ${seq.id} has running instances, use --force to kill those.`);
+                        continue;
+                    }
+                    // maybe we should stop
+                    displayMessage(`Killing instances of sequence ${seq.id}`);
+                    await Promise.all(
+                        seq.instances.map(async (id) => {
+                            if (lastInstanceId === id) sessionConfig.setLastInstanceId("");
+
+                            return getInstance(id).kill();
```
a-tylenda, line 166:
```suggestion
                            return getInstance(id).kill();
                            defer(1000);
```

maybe we could give some time for Sequence to stop working?

packages/cli/src/lib/commands/sequence.ts
```diff
@@ -140,15 +173,43 @@ export const sequence: CommandDefinition = (program) => {
...
-            .description("delete multiple sequences on actual selected hub")
-            .action(() => {
-            // FIXME: implement me
-                throw new Error("Implement me");
-            });
+    sequenceCmd
+        .command("prune")
+        // .option("--all")
+        // .option("--filter")
+        .option("-f,--force", "Removes also active sequences")
```
a-tylenda, line 180:
```suggestion
        .option("-f,--force", "Removes also active sequences")
        .description("remove all Sequences from the current scope (use with caution)")
        
```

a-tylenda:
verification complete:

image

PR #432, opened 2022-04-13
Fix and test passing args to sequence through CLI.

### LINKED ISSUE #431: Cannot pass args with CLI when starting sequence
The CLI expects the JSON string as the value for `--args` in the `si seq start` command. It's not parsed anywhere later so the Runner receives a string and the arguments are not correctly dispatched to the sequence.

pietrzakacper:
This fixes #431 bug and is needed for the documentation to be accurate in #420

### DIFF
--- a/packages/cli/src/lib/commands/sequence.ts
+++ b/packages/cli/src/lib/commands/sequence.ts
@@ -94,8 +94,12 @@ export const sequence: CommandDefinition = (program) => {
             .option("--input-topic <string>", "topic to which the input stream should be routed")
             .option("--args <json-string>", "arguments to be passed to first function in Sequence")
             .description("start the sequence with or without given arguments")
-            .action(async (id, { configFile, configString, outputTopic, inputTopic, args }) =>
-                startSequence(id, { configFile, configString, args, outputTopic, inputTopic }));
+            .action(async (id, { configFile, configString, outputTopic, inputTopic, args: argsStr }) => {
+                const args = argsStr ? JSON.parse(argsStr) : [];
+
+                await startSequence(id, { configFile, configString, args, outputTopic, inputTopic });
+            }
+            );
     else
         sequenceCmd
             .command("start")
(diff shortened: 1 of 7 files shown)


packages/cli/src/lib/commands/sequence.ts
```diff
@@ -94,8 +94,12 @@ export const sequence: CommandDefinition = (program) => {
             .option("--input-topic <string>", "topic to which the input stream should be routed")
             .option("--args <json-string>", "arguments to be passed to first function in Sequence")
             .description("start the sequence with or without given arguments")
-            .action(async (id, { configFile, configString, outputTopic, inputTopic, args }) =>
-                startSequence(id, { configFile, configString, args, outputTopic, inputTopic }));
+            .action(async (id, { configFile, configString, outputTopic, inputTopic, args: argsStr }) => {
+                const args = argsStr ? JSON.parse(argsStr) : [];
```
ErykSol, line 98: unhandled exception from parse

packages/cli/src/lib/commands/sequence.ts
```diff
@@ -104,8 +108,12 @@ export const sequence: CommandDefinition = (program) => {
             .option("-s, --config-string <json-string>", "configuration in JSON format to be passed to instance context")
             .option("--args <json-string>", "arguments to be passed to first function in Sequence")
             .description("start the sequence with or without given arguments")
-            .action(async (id, { configFile, configString, args }) =>
-                startSequence(id, { configFile, configString, args }));
+            .action(async (id, { configFile, configString, args: argsStr }) => {
+                const args = argsStr ? JSON.parse(argsStr) : [];
```
ErykSol, line 112: unhandled exception from parse

ErykSol (changes requested):

alicja-gruzdz (approved):

packages/cli/src/lib/commands/sequence.ts
pietrzakacper: fixed

packages/cli/src/lib/commands/sequence.ts
pietrzakacper: fixed

ErykSol (approved):

PR #433, opened 2022-04-13
Release 0.21

jan-warchol:
- New command `si prune` for removing old sequences
- Minor CLI bugfixes


jan-warchol:
Not this time.

PR #435, opened 2022-04-13
Fix data missing in handshake return

Tatarinho:


jan-warchol (commented):
Is it possible that `data` will be `None` or some other value that will crash the `if ... in data` checks?

jan-warchol (approved):

alicja-gruzdz (approved):

PR #438, opened 2022-04-13
Add getConfig method to HostClient

jan-warchol:
The endpoint was added in 212be1d15ccb433327d958b4127ed582fcce1d78.

### DIFF
--- a/packages/api-client/src/host-client.ts
+++ b/packages/api-client/src/host-client.ts
@@ -107,6 +107,15 @@ export class HostClient implements ClientProvider {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


jan-warchol:
@MichalCz am I doing this right?

patuwwy (approved):

patuwwy:
What about CLI ?

jan-warchol:
> What about CLI ? @agne6ka

I _think_ this is something we don't plan to make available through CLI, so I'll merge this PR, but if I'm wrong, let me know and I'll make the additions.

PR #439, opened 2022-04-13
Rename 'python-runner' to 'runner-py'

jan-warchol:
That's how we're naming our docker image, and we want package name and
image name be consistent. Also, this way 'runner-py' will be next to
'runner' when sorted lexicographicaly.


jan-warchol:
@patuwwy how this should be handled in NPM? A new package?

jan-warchol:
It seems we have a chicken&egg problem here, as one module depends on runner and either in case of NPM or running from source the naming will be wrong..

MichalCz:
@jan-warchol you just forgot to add the `runner-py` to dependencies of `adapters` so it couldn't find it and it fetched it from npm.

MichalCz:
Fixed, but there are other errors after rebase. I'll fix after merges in #485

PR #450, opened 2022-04-14
Enable closing HTTP connection in API Clients

pietrzakacper:
Sometimes we want to stop reading the response stream e.g. output. For the node programs to properly exit we also need to close the underlying TCP connection when we no longer want to read the response stream.

AbortController should be available both in Node (>15, so the polyfill is needed) and in browsers for a couple of years now.


MichalCz (approved):
Dep ok, lgtm.

patuwwy (changes requested):
https://github.com/node-fetch/node-fetch#request-cancellation-with-abortsignal

patuwwy (approved):

PR #454, opened 2022-04-15
Select first healthy hub as default in CLI

### LINKED ISSUE #449: CLI: Add healthy check when setting the platform defaults
When the CLI platform defaults are set (the Space and the Hub), the CLI should check if there is any Hub with the `healthy:true` in the Hub list. Then it could take e.g. the first healthy Hub and set it as the default one. We would need to solve this issue for the Manager too, but I will focus here only on the Hubs.

Currently, the CLI takes the first element from the retrieved spaces and the Hubs arrays: https://github.com/scramjetorg/transform-hub/blob/devel/packages/cli/src/lib/platform/common.ts#L58
Alternately, the CLI could at least signal that the set Hub is unhealthy and suggest listing the Hubs to check if there are any responsive ones.

@daro1337 and I had an issue that the CLI set unhealthy Hub as the default one and the response for any command (e.g. `si seq ls`)  was just hanging without any error.

pietrzakacper:
This fixes #449. When there multiple hubs in user space, we want to select the first healthy one as the default one.

### DIFF
--- a/packages/cli/src/lib/platform/common.ts
+++ b/packages/cli/src/lib/platform/common.ts
@@ -50,16 +50,21 @@ export const setPlatformDefaults = async () => {
 
         if (!managers.length) return false;
 
-        const managerClient = middlewareClient.getManagerClient(managers[0].id);
+        const selectedManager = managers[0]
+
+        const managerClient = middlewareClient.getManagerClient(selectedManager.id);
         const hosts = await managerClient.getHosts();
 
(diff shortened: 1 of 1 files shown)


alicja-gruzdz (approved):
thanks Kacper, it will make using CLI easier.
Verify (after removing the CLI configs and running the CLI for the first time following that):
Zrzut ekranu z 2022-04-15 16-01-26

PR #455, opened 2022-04-15
[Bugfix] Parse seq args in CLI "si seq deploy"

a-tylenda:
Changes done by @pietrzakacper in PR [#432](https://github.com/scramjetorg/transform-hub/pull/432) also needed to be implemented in `si seq deploy --args '[]'`

test added: `NO_HOST=1 yarn test:bdd --name="E2E-010 TC-029"`


alicja-gruzdz (approved):
Good catch & fix!

patuwwy (approved):

PR #458, opened 2022-04-21
The message type + codes for the audit stream

alicja-gruzdz:
The audit stream carries information on all operations performed in the platform.

The message codes included are for the basic operations; 
- operations on Instance,
- operations on Sequence,
- operations on Topics,
- internal system operations.

Suggestions on the improvements of the message code created are very welcome. Perhaps for the codes we should use string and create a hash from the request method name (can be digits) + the part of URL (after removing the base and the IDs from the URL). 
The current codes are based on @patuwwy suggestion to have the first digit indicating a category of operations (Instance, Sequence, etc.) and the second digit for the method. However, the last digits would still need to be known. 

There is only one message type proposed with all required attributes listed without nesting. It's not the neatest modelling, but otherwise we would need to maintain a long list of various combinations of message codes and attributes.

[EDIT]: After debugging the Auditor service with @patuwwy I adopted his status codes convention which can be found in https://github.com/scramjetorg/transform-hub/pull/461/files#diff-10ce2297b522d8ec475f58fc3b0c9022d899a48ff62cb9430c28f503ccf36389

### DIFF
--- a/packages/symbols/src/op-record-code.ts
+++ b/packages/symbols/src/op-record-code.ts
@@ -1,0 +1,3 @@
+export enum OpRecordCode {
+
+}
--- a/packages/symbols/src/op-record-code.ts
+++ b/packages/symbols/src/op-record-code.ts
@@ -24,7 +24,7 @@ export enum OpRecordCode {
 
     /**
      * The message codes related to the operation on a Sequence.
-     * They are trigger in response for the user operations performed by the API.
+     * They are triggered in response to the user operations performed by the API.
      */
 
     GET_SEQUENCE = 11200,
--- a/packages/symbols/src/op-record-code.ts
+++ b/packages/symbols/src/op-record-code.ts
@@ -1,3 +1,60 @@
 export enum OpRecordCode {
 
+    /**
+     * The message codes related to the operation on an Instance.
+     * They are trigger in response for the user operations performed by the API.
+     */
+
+    GET_INSTANCE = 8000,
+    GET_INSTANCE_STDOUT = 8001,
+    GET_INSTANCE_STDERR = 8002,
+    GET_INSTANCE_STDIN = 8003,
+    GET_INSTANCE_LOG = 8004,
+    GET_INSTANCE_MONITORING = 8005,
+    GET_INSTANCE_OUTPUT = 8006,
+    GET_INSTANCE_HEALTH = 8007,
+    GET_INSTANCE_EVENTS = 8008,
+    GET_INSTANCE_EVENT = 8009,
+    GET_INSTANCE_EVENT_ONCE = 8010,
+    POST_INSTANCE_INPUT = 8100,
+    POST_INSTANCE_MONITORING_RATE = 8101,
+    POST_INSTANCE_EVENT = 8102,
+    POST_STOP_INSTANCE = 8103,
+    POST_KILL_INSTANCE = 8104,
+
+    /**
+     * The message codes related to the operation on a Sequence.
+     * They are trigger in response for the user operations performed by the API.
+     */
+
+    GET_SEQUENCE = 9000,
+    DELETE_SEQUENCE = 9200,
+    POST_SEQUENCE = 9100,
+    POST_START_INSTANCE = 9101,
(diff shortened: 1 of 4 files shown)


alicja-gruzdz:
After the conversation with @patuwwy we decided to focus only on the API operation registration for now. When all is working we will be extended the message in attributes and codes as they are required.

packages/symbols/src/op-record-code.ts
```diff
@@ -0,0 +1,60 @@
...
+    GET_INSTANCE_EVENT_ONCE = 8010,
+    POST_INSTANCE_INPUT = 8100,
+    POST_INSTANCE_MONITORING_RATE = 8101,
+    POST_INSTANCE_EVENT = 8102,
+    POST_STOP_INSTANCE = 8103,
+    POST_KILL_INSTANCE = 8104,
+
+    /**
+     * The message codes related to the operation on a Sequence.
+     * They are trigger in response for the user operations performed by the API.
```
a-tylenda, line 27:
```suggestion
     * They are triggered in response to the user operations performed by the API.
```

packages/symbols/src/op-record-code.ts
```diff
@@ -0,0 +1,60 @@
+export enum OpRecordCode {
+
+    /**
+     * The message codes related to the operation on an Instance.
+     * They are trigger in response for the user operations performed by the API.
```
a-tylenda, line 5:
```suggestion
     * They are triggered in response to the user operations performed by the API.
```

packages/symbols/src/op-record-code.ts
a-tylenda, line 27:
```suggestion
     * They are triggered in response to the user operations performed by the API.
```

packages/symbols/src/op-record-code.ts
```diff
@@ -0,0 +1,60 @@
...
+     */
+
+    GET_SEQUENCE = 9000,
+    DELETE_SEQUENCE = 9200,
+    POST_SEQUENCE = 9100,
+    POST_START_INSTANCE = 9101,
+
+    /**
+     * The message codes related to the operation on Topics.
+     * They are trigger in response for the user operations performed by the API.
```
a-tylenda, line 37:
```suggestion
     * They are triggered in response to the user operations performed by the API.
```

a-tylenda (approved):

packages/symbols/src/op-state.ts
```diff
@@ -0,0 +1,6 @@
+export enum OpState {
+
```
patuwwy, line 2:
```suggestion
```

packages/types/src/messages/op-record.ts
```diff
@@ -0,0 +1,47 @@
+import { OpRecordCode } from "@scramjet/symbols";
+
+export type OpRecord = {
+
```
patuwwy, line 4:
```suggestion
```

packages/types/src/messages/op-record.ts
```diff
@@ -0,0 +1,47 @@
...
+    /**
+    * The operation state from the OpState enumeration.
+    */
+    opState: string;
+
+    /**
+    * The generated unique request ID for operations coming from the API.
+    */
+
```
patuwwy, line 18:
```suggestion
```

packages/types/src/messages/op-record.ts
```diff
@@ -0,0 +1,47 @@
...
+    /**
+    * The data received in the request stream
+    */
+    rx?: number;
+
+    /**
+    * An instance of the object which the operation concerns, e.g. Host.
+    */
+    objectId: string;
+
```
patuwwy, line 45:
```suggestion
```

alicja-gruzdz: thanks :)

packages/symbols/src/op-record-code.ts
alicja-gruzdz: thanks Aga!

MichalCz (approved):

PR #462, opened 2022-04-22
Change source of framework-py to official pypi release

Tatarinho:


f1ames (approved):
CI is green, which means it works! :tada: Good job :+1:

PR #463, opened 2022-04-22
Build Docker image based on package.json

Tatarinho:
Basic, primitive solution. A lot of things to discuss.

Process adapter didn't change.
Docker adapter returns image based on SequenceConfig, build if it's necessary

Didn't tested with k8s. Probably doesn't work.


packages/adapters/src/dockerode-docker-helper.ts
```diff
@@ -198,6 +199,39 @@ export class DockerodeDockerHelper implements IDockerHelper {
...
+        //TODO: cannot tag image with dockerode, fix?
+        // const buildStream = await this.dockerode.buildImage({
+        //     context: path.resolve(__dirname, "../../../packages/python-runner"),
+        //     src: ["Dockerfile"]
+        // }, { t: `scramjetorg/runner-py:${pythonVersion}` }
+        // );
+
+        const context = " .";
+        const buildArgs = ` --build-arg IMAGE=${version}`;
+        const command = "docker build -t " + name + buildArgs + " -f " + dockerfilePath + context;
```
MichalCz, line 214:
```suggestion
        const command = `docker build -t ${name}${buildArgs} -f ${dockerfilePath}${context}`;
```

PR #464, opened 2022-04-25
SCP-3531 proper PID handling within containers

daro1337:
[SCP-3531] Use an init system - tini


MichalCz (approved):

MichalCz:
Let's merge. Tests have passed so this seems to work well.

@daro1337 is this confirmed to work well?

stokowski (approved):
lgtm

PR #466, opened 2022-04-25
Testing: API server not allowing half open connections

alicja-gruzdz:
Just experimenting with the option to see what effects it will have on the BDD tests.
feat. @patuwwy


alicja-gruzdz:
Conclusion: for certain requests passed from the Manager to the Host (e.g. start Sequence) it was required to allow half open.

PR #467, opened 2022-04-26
Include the original API response status code and message in the api-client error type

alicja-gruzdz:
Before we publish the list of Scramjet Transform Hub API error codes it would be helpful to have the same error codes and messages returned by the API and the API-Client.

The things that still need to be completed:

- [x] Following @patuwwy  suggestion: modify the `@api-server` to NOT set chunked encoding when the error is thrown.
Then we can easier read the @api-client's `ClientError` `body` property and include the original API error information in the `ClientError`. 
[EDIT] I finally ended up using @pietrzakacper tip and accessed the original error with https://www.npmjs.com/package/node-fetch#bodyjson
- [x] Remove the new empty `@constants` package. It is not required since we will place errors in the `@types` and the error codes in the `@symbols`.
The response status codes which the package was meant for will be from the https://www.npmjs.com/package/http-status-codes
- [x] Add error codes enum to @symbols
- [x] Add APIErrorMessage to @types

Nice to have:
- [ ] Modify the basic errors returned in the STH.


patuwwy (approved):

packages/client-utils/src/client-error.ts
```diff
@@ -63,16 +64,17 @@ export class ClientError extends Error {
...
-                if (+error.code >= 500) return new this("SERVER_ERROR", error, message, source);
-                if (+error.code >= 400) return new this("REQUEST_ERROR", error, message, source);
-                return new this("UNKNOWN_ERROR", error, `Response code is "${error.code}"`, source);
+                if (code === "400") return new this("BAD_PARAMETERS", error, message, source, code);
+                if (code === "401") return new this("NEED_AUTHENTICATION", error, message, source, code);
+                if (code === "403") return new this("NOT_AUTHORIZED", error, message, source, code);
+                if (code === "404") return new this("NOT_FOUND", error, message, source, code);
+                if (code === "410") return new this("GONE", error, message, source, code);
+                if (code === "422") return new this("UNPROCESSABLE_ENTITY", error, message, source, code);
+                if (code === "507") return new this("INSUFFICIENT_RESOURCES", error, message, source, code);
```
a-tylenda, line 73:
```suggestion
                if (code === "419") return new this("INSUFFICIENT_RESOURCES", error, message, source, code);
```

alicja-gruzdz: thanks Aga, fixed :)

MichalCz (approved):

jan-warchol:
So, merge? :)

alicja-gruzdz:
> So, merge? :)

ok, here I go :)

PR #471, opened 2022-04-26
Handle async generator in runner-py

Tatarinho:
As part of this PR:
- [x] Handle async generator in runner-py
- [x] Add sample with async generator and BDD test for it
- [x] Add building runner-py image on CI

### DIFF
--- a/packages/python-runner/runner.py
+++ b/packages/python-runner/runner.py
@@ -6,6 +6,7 @@ import json
 from pyee import EventEmitter
 import importlib.util
 from io import DEFAULT_BUFFER_SIZE as CHUNK_SIZE
+import types
 
 from scramjet.streams import Stream
 from logging_setup import LoggingSetup
@@ -184,7 +185,6 @@ class Runner:
         # switch to sequence dir so that relative paths will work
         os.chdir(os.path.dirname(self.seq_path))
 
-
     async def run_instance(self, config, args):
         context = AppContext(self, config)
         input_stream = Stream()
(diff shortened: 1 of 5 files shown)


packages/python-runner/runner.py
```diff
@@ -205,7 +205,9 @@ async def run_instance(self, config, args):
             self.logger.info(f'Sending PANG with {consumes}')
             send_encoded_msg(monitoring, msg_codes.PANG, consumes)
 
-        if asyncio.iscoroutine(result):
+        if isinstance(result, types.AsyncGeneratorType):
+            result = Stream.read_from(result)
```
f1ames, line 209: This is basically very similar to [what we have in JS runner](https://github.com/scramjetorg/transform-hub/blob/devel/packages/runner/src/runner.ts#L476-L480) so it should align runner behaviors :+1:

python/reference-apps/python-gen-async/main.py
```diff
@@ -0,0 +1,7 @@
+async def save_entry_to_db(id: str) -> None:
```
f1ames, line 1: Maybe a test which run this sequence? Or you think it's not necessary?

f1ames (dismissed):
Looks good :+1: 

Just one suggestion regarding tests (see comment).

python/reference-apps/python-gen-async/main.py
Tatarinho: I can write one simple test I think

MichalCz (approved):

PR #473, opened 2022-04-26
Audit host,csics heartbeat on audit request

patuwwy:


alicja-gruzdz (approved):
Tested:
- heartbeat from the Host
- heartbeat from the Instance: starting and stopping the Instances while checking if their heartbeat was added or removed from the audit stream.
Zrzut ekranu z 2022-04-26 19-21-48

a-tylenda (approved):

PR #479, opened 2022-05-04
Settings for cucumber

a-tylenda:
I added some basic settings for cucumber which enable:
- Auto-parsing of feature steps from paths, provided in settings.json (to follow the step definition use "Ctrl+click" over the feature step you want to follow)
- Autocompletion of steps

it will work only when this VS Code extension is installed: [Given When Then](https://marketplace.visualstudio.com/items?itemName=alexkrechik.cucumberautocomplete)


alicja-gruzdz (approved):
maybe it's trivial, but it's very useful :+1: 
I'd add info about it to readme.

a-tylenda:
> 

done

MichalCz (approved):

PR #483, opened 2022-05-05
Fix python deps path installation

jan-warchol:
Somehow we ended up with two different paths, __pydeps__ and
__pypackages__. Since only one was included in PYTHONPATH, the other
(used when running from source) didn't work.

Also, include original PYTHONPATH only if it's defined and remove
redundant path.resolve wrap.

https://app.clickup.com/t/24308805/SCP-3676


Tatarinho (approved):

packages/adapters/src/process-instance-adapter.ts
```diff
@@ -78,12 +78,16 @@ IComponent {
...
         let pythonpath = path.resolve(
-            __dirname,
-            path.resolve(__dirname, require.resolve("@scramjet/python-runner"), "..")
+            __dirname, require.resolve("@scramjet/python-runner"), "../__pypackages__"
         );
 
-        pythonpath += `:${process.env.PYTHONPATH}:${sequenceDir}/__pypackages__`;
+        if (process.env.PYTHONPATH) pythonpath += `:${process.env.PYTHONPATH}`;
+
+        pythonpath += `:${sequenceDir}/__pypackages__`;
```
MichalCz, line 90: How about docker and kube adapters? Won't this change break that?

MichalCz (changes requested):

.gitignore
```diff
@@ -137,7 +137,7 @@ aaa/
 
 **/__pycache__/
 
-**/__pydeps__/**
+**/__pypackages__/**
```
MichalCz, line 140: @jan-warchol but I meant this - why do we change this and won't this break existing `__pydeps__` dirs

MichalCz (changes requested):

alicja-gruzdz:
Any verify needed?

jan-warchol:
> Any verify needed?

Yes, if you could check this with docker & k8s adapters to confirm that everything works it would be great.

.gitignore
jan-warchol: Ok, I assume I can mark this as resolved?

f1ames (changes requested):
Hmmm, first I tried with Python sequence from `python/reference-apps` but it resulted in the following error:

> DEVELOPMENT=1 yarn start -P 8000 --runtime-adapter=process

image

This looks like sequence code is outdated? :thinking: 

So then I took `hello-py` sequence from `platform-tests` repo which seems to work fine.

## With process adapter

It works with dist version:

> DEVELOPMENT=1 yarn start -P 8000 --runtime-adapter=process

image

but errors with dev version (which fixing was the purpose of this PR AFAIU):

> DEVELOPMENT=1 yarn start:dev -P 8000 --runtime-adapter=process

image

## Docker adapter

With docker adapter for both `dist` and `dev` versions work fine.

## K8s

Checking in progress...

f1ames (approved):
After talking F2F with @jan-warchol it seems there was some cache issue. Some magic with `git clean -fdx` helped. I tested it with both sequences I initially used (since `python/reference-apps/python-alice` started to work too).

## `process` dev

* `python-alice` :heavy_check_mark: 
* `hello-py` :heavy_check_mark: 

## `process` dist

* `python-alice` :heavy_check_mark: 
* `hello-py` :heavy_check_mark: 

## `docker` dev

* `python-alice` :heavy_check_mark: 
* `hello-py` :heavy_check_mark: 

## `docker` dist

* `python-alice` :heavy_check_mark: 
* `hello-py` :heavy_check_mark: 

---

However, I have noticed one issue (looks like not related to this PR, but worth checking). The following scenario fails:

1. Send sequence (`python-alice`).
1. Start it and wait to be finished.
1. Send another sequence (`hello-py`).
1. Start it and wait to be finished.

The last step fails and it seems sequence is not started, it just hangs after receiving `PANG` message (both `process` and `docker` adapters), see below the last thing logged and then it just hangs:

With `process`:

image

With `docker`:

image

jan-warchol:
Thanks for verifying!

> However, I have noticed one issue (looks like not related to this PR, but worth checking). The following scenario fails:
> 
> 1. Send sequence (`python-alice`).
> 2. Start it and wait to be finished.
> 3. Send another sequence (`hello-py`).
> 4. Start it and wait to be finished.
> 
> The last step fails and it seems sequence is not started, it just hangs after receiving `PANG` message (both `process` and `docker` adapters), see below the last thing logged and then it just hangs:
> 
> With `process`:
> 
> image
> 
> With `docker`:
> 
> image

This sequence just waits for some input. I see the same in logs when I start it, and when I send something to the input I can get appropriate output.

jan-warchol:
Strange indeed. Does it work when you send input? (open `si inst output` and in another terminal `si inst input`, write some text and press enter)

PR #484, opened 2022-05-05
Capture runner logs and show in case of failure

jan-warchol:
Currently if the runner process crashes before it connects to the host
(e.g. because an import fails on missing dependency), we don't see
stdout/err unless DEVELOPMENT is set. This makes it hard to debug some
situations (e.g. https://app.clickup.com/t/24308805/SCP-3676).

The new approach is to save all output to a file and show it in case of
failure.

We could also:
- consider saving the logs only for some time
- displaying only last N lines in case of failure
- try to detect when runner tcp connection is established and stop capturing logs then
- somehow give logfiles more meaningful names (with PID/container ID?)

What do you think?

### DIFF
--- a/packages/adapters/src/process-instance-adapter.ts
+++ b/packages/adapters/src/process-instance-adapter.ts
@@ -10,6 +10,7 @@ import {
     SequenceConfig
 } from "@scramjet/types";
 import { ChildProcess, spawn } from "child_process";
+import { createReadStream, createWriteStream } from "fs";
 
 import path from "path";
 
@@ -125,9 +126,16 @@ IComponent {
             }
         });
 
+        const runnerLogFile = path.resolve(__dirname, "runner-startup.log");
+
         if (development()) {
             runnerProcess.stdout.pipe(process.stdout);
             runnerProcess.stderr.pipe(process.stderr);
+        } else {
+            const runnerLog = createWriteStream(runnerLogFile);
+
+            runnerProcess.stdout.pipe(runnerLog);
+            runnerProcess.stderr.pipe(runnerLog);
         }
 
         this.logger.trace("Runner process is running", runnerProcess.pid);
@@ -149,6 +157,11 @@ IComponent {
 
         if (statusCode > 0) {
             this.logger.debug("Process returned non-zero status code", statusCode);
+
+            if (!development()) {
+                this.logger.debug("Forwarding stdout & stderr from runner:");
+                createReadStream(runnerLogFile).pipe(process.stderr);
+            }
         }
 
         return statusCode;


alicja-gruzdz:
> What do you think?

I think it would be wonderful to have any logs that would assist in these annoying Runner crashes.
The third solution seems to be the most intelligent but not the simplest.
Honestly, for me right now even the N lines of logs would be fine.

In general this solution would be very useful in the Beta and I'm thinking how we could make these logs available to the user there? CSIController lifespan is limited, unless we extend it and agree on a fixed amount of time it will be not destroyed after the Runner exits.

alicja-gruzdz:
> * somehow give logfiles more meaningful names (with PID/container ID?)

yep, and timestamp - I know there's file creation time but in case of moving the file :)

packages/adapters/src/process-instance-adapter.ts
```diff
@@ -149,6 +157,11 @@ IComponent {
 
         if (statusCode > 0) {
             this.logger.debug("Process returned non-zero status code", statusCode);
+
+            if (!development()) {
+                this.logger.debug("Forwarding stdout & stderr from runner:");
+                createReadStream(runnerLogFile).pipe(process.stderr);
```
alicja-gruzdz, line 163: btw: do you need to later (after the CSI Controller exits) unpipe the createReadStream(runnerLogFile).pipe(process.stderr); ?

alicja-gruzdz (commented):
btw: do you need to later unpipe createReadStream(runnerLogFile).pipe(process.stderr); ?

alicja-gruzdz (approved):
I generally agree with the direction. We still have to answer a couple of questions, but I'm glad someone is looking into this issue.

MichalCz:
Actually this should be done in runner - any data could not be sent over these channels in case of a network failure. I have a PR, so we can merge this, but I'll rewrite this a little bit later... or just not... anything I do is "ETA unknown" as we know... ;)

alicja-gruzdz:
> Actually this should be done in runner - any data could not be sent over these channels in case of a network failure. I have a PR, so we can merge this, but I'll rewrite this a little bit later... or just not... anything I do is "ETA unknown" as we know... ;)

Indeed, I forgot about your https://github.com/scramjetorg/transform-hub/pull/408
So does it mean that you agree to merge this for now, but you would like to correct it later in the #408?

packages/adapters/src/process-instance-adapter.ts
MichalCz: Yes, it would seem so... I'll check this, as we've got some hanging pipes in multiple locations and I see warnings about listeners piling up in a couple places.

PR #488, opened 2022-05-09
Explicitly define adapter as one of 'docker', 'process', 'kubernetes'

jan-warchol:
This will be handy when using the value of this attribute to get other
attributes, like config[config.runtimeAdapter].runnerImages.python3


jan-warchol:
Too much work for too little gain.

PR #490, opened 2022-05-10
CLI: Do not display a stream response when sending input to the Instance

alicja-gruzdz:
A tiny fix for the CLI displaying a PassThrough stream response when sending input to the Instance.

Before:
Screenshot from 2022-05-10 14-19-48

After:
- With debug:
Screenshot from 2022-05-10 14-20-15
- Without debug:
Screenshot from 2022-05-10 14-20-43


patuwwy (approved):
thx

PR #491, opened 2022-05-11
Handle Runner failure and Sequence errors on Sequence start

pietrzakacper:
This addresses both: https://app.clickup.com/t/24308805/SCP-3756 and https://app.clickup.com/t/24308805/SCP-3757

Now starting a sequence through API or CLI should properly exit with error in the following cases:
1. Runner process exited with non zero code before receiving PANG message
2. Sequence entrypoint file throws when importing it (e.g. syntax errors)
3. Sequence is an empty array 

### 1. Crashing runner process
1. Add somewhere in global scope of `packages/runner/src/runner.ts`
```ts
throw new Error("OOOPS");
``` 
2. Run STH: 
```bash
yarn start:dev
```
3. In a new terminal run CLI:
```
npx ts-node packages/cli/src/bin/index.ts seq send ./packages/reference-apps/hello-alice-out.tar.gz
npx ts-node packages/cli/src/bin/index.ts seq start -
```
4. You should get similar output:
image

### 2. Sequence entrypoint file is invalid
1. Add somewhere in global scope of some ref-app and then build it
```ts
throw new Error("OOOPS");
``` 
2. Run STH: 
```bash
yarn start:dev
```
3. In a new terminal run CLI:
```bash
npx ts-node packages/reference-apps/<path-to-your-sequence>
npx ts-node packages/cli/src/bin/index.ts seq start -
```
4. You should get similar output:
image

### 3. Empty sequence
Do the same as in 2nd test, just export an empty array:
```ts
const app: any[] = []
export default app;
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "lint:dedupe": "scripts/dedupe.js",
     "start": "node dist/sth/bin/hub.js",
     "start:dev": "ts-node packages/sth/src/bin/hub.ts",
+    "start:dev:cli": "ts-node packages/cli/src/bin/index.ts",
     "install:clean": "yarn clean && yarn clean:modules && yarn install",
     "postinstall": "lerna link",
     "prepare": "npx husky install",
--- a/packages/api-client/src/sequence-client.ts
+++ b/packages/api-client/src/sequence-client.ts
@@ -59,7 +59,7 @@ export class SequenceClient {
         );
 
         if (response.result === "error") {
-            return Promise.reject({ error: response.error });
+            return Promise.reject(response.error);
         }
 
         return InstanceClient.from(response.id, this.host);
--- a/packages/cli/src/lib/commands/sequence.ts
+++ b/packages/cli/src/lib/commands/sequence.ts
@@ -7,7 +7,7 @@ import { CommandDefinition } from "../../types";
 import { isDevelopment } from "../../utils/isDevelopment";
 import { getHostClient, getInstance, getReadStreamFromFile, packAction } from "../common";
 import { getPackagePath, getSequenceId, sessionConfig } from "../config";
-import { displayEntity, displayMessage, displayObject } from "../output";
+import { displayEntity, displayError, displayMessage, displayObject } from "../output";
 
 const sendPackage = async (sequencePackage: string) => {
     const seq = await getHostClient().sendSequence(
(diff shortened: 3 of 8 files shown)


alicja-gruzdz:
@jan-warchol when you have a free moment please take a look at it, it solves two of the issues you have reported.

jan-warchol:
> @jan-warchol when you have a free moment please take a look at it, it solves two of the issues you have reported.

Thanks, I almost missed it! I'll check how it works with python runner.

jan-warchol:
Ok, so the first case (crashing runner) works nice :+1: , but in the second case there is no difference in case of python runner (meaning that I get a `200 OK` while the sequence crashes). I suppose this is because python runner loads the sequence after sending the PANG.

(To reproduce, just add `raise SyntaxError` to some python sequence, send it and try starting.)

I guess the question is: does it indicate a problem with python runner, or should we change the detection method?

alicja-gruzdz:
> I guess the question is: does it indicate a problem with python runner, or should we change the detection method?

Thanks for checking it. Indeed, you're right. It looks like we Node and Python Runner behave differently in this regard. From what I see in Python Runner the PANG is sent before detecting the Sequence in https://github.com/scramjetorg/transform-hub/blob/4b9aa51bd08ca552e2fab53d3d478f83be1ecb75/packages/python-runner/runner.py#L175. 
In Node the PANG goes out of the Runner only if it passes the detection in 
https://github.com/scramjetorg/transform-hub/blob/devel/packages/runner/src/runner.ts#L407

Please correct me if I'm wrong.
If so, it would be better to modify it in the Python Runner to behave like in the Node Runner.

From what @MichalCz mentioned this week if the PANG goes out of the Runner it means the required Sequence's structure is valid and we do not need to raise an API error on Sequence start.

alicja-gruzdz:
Thanks Kacper for taking this herd of fixes.

Verify:
1. As long as the error is thrown before PANG the API error is returned :heavy_check_mark: 
before_getSequence

2. Incorrect Sequence main file :heavy_check_mark: 
Sequence_entrypoint_err

3. Empty Sequence :heavy_check_mark: 
empty

alicja-gruzdz (approved):

pietrzakacper:
@jan-warchol I'm merging as this PR already brings some value. I see two options:
1. We load the sequence before the pang in python-runner
2. We have a special kind of monitoring message that is specifically sent after the sequence is loaded and verified (e.g. checked for correct export structure). The runner should send that message or quit.

Either way we have to modify the python runner in order to check to do any error handling in send sequence endpoint.

MichalCz:
Yes, in my view PANG means that we've got everything working and the syntax is ok.

PR #494, opened 2022-05-12
Move (almost all) reference apps to separate repo

jan-warchol:
* Remove (almost all) reference apps

  They have their own repo, https://github.com/scramjetorg/reference-apps
  However, a few of them are quite tightly coupled to BDD tests code, so
  for now I'm keeping them - they should be decoupled later.

* Add script for downloading prebuilt refapps (for local testing)

* CI: replace building refapps with downloading prebuilt ones

  We now have a separate repo for reference apps, with tar.gz files
  available as released artifacts. This is much faster than building them
  every time.

Basically, most of this PR is just deleting files. Fortunately GH folds their previews, so you can simply focus on reviewing visible code changes.


Tatarinho (approved):

bdd/step-definitions/e2e/host-steps.ts
```diff
@@ -221,10 +221,16 @@ When("wait for {string} ms", async (timeoutMs: number) => {
 });
 
 When("sequence {string} loaded", { timeout: 50000 }, async function(this: CustomWorld, packagePath: string) {
```
Tatarinho, line 223: two similar steps, maybe it is good idea to remove it?

jan-warchol: To be honest, I really would like not deal with them in this PR. I've tried untangling step definitions in the past and it usually turned out that they are interconnected in some weird ways - changing one requires changing some global variables, which affect other steps, and so on...

So, while it seems trivial, I'd say there's a 40% chance that it will eat up half a day.

a-tylenda (approved):

bdd/step-definitions/e2e/host-steps.ts
a-tylenda: I agree with @jan-warchol , I also tried once to unify similar steps and failed, but sooner or later it will have to be done I guess.

PR #496, opened 2022-05-13
Remove all cucumber scenarios that are not run by CI

jan-warchol:
We should not keep unused test cases. What's more important, we should
not have test cases that are ran locally but not on CI (or vice versa).
It's important for test reliability that running tests locally and on CI
will test the same things.

This PR simply deletes all test cases that were not enabled on CI. However, if you think that some of them are valuable, there are other options:
* we could comment them out
* we could put it in a separate file that is not accessed by cucumber.

What do you think? Please comment.


MichalCz (approved):

bdd/features/e2e/E2E-010-cli.feature
```diff
@@ -63,19 +63,6 @@ This feature checks CLI functionalities
...
-        Given host is running
-        Then I set json format
-        Then I use apiUrl in config
-        When I execute CLI with "seq send ../packages/reference-apps/hello-alice-out.tar.gz" arguments
-        Then I get Sequence id
-        Then I start Sequence
-        And I get Instance id
-        Then I get Instance log
-        And host is still running
-
```
a-tylenda, line 78: This test refers to CLI command `si inst log <id>`. I think it is nice to have. It works after small refactor:
```
    @ci @cli
    Scenario: E2E-010 TC-006 Get log from Instance
        Given host is running
        Then I set json format
        Then I use apiUrl in config
        When I execute CLI with "seq send ../packages/reference-apps/inert-function.tar.gz" arguments
        Then I get Sequence id
        Then I start Sequence
        And I get Instance id
        When I execute CLI with "inst log -" arguments
        And host is still running
```
Executing the log test locally:

image

a-tylenda: I will try to revive it in a separate PR, to make sure it works on CI.

a-tylenda (approved):
Scenarios are removed, what about redundant step definitions code which is left in `*.ts` files? Will it be removed in a separate PR?

bdd/features/e2e/E2E-010-cli.feature
jan-warchol: Good idea!

jan-warchol:
> Scenarios are removed, what about redundant step definitions code which is left in `*.ts` files? Will it be removed in a separate PR?

Good point! I'll try to remove them in this PR.

jan-warchol:
Removed.

Issue #505, opened 2022-05-18
[CLI] "si hub ls" does not return output in json format


daro1337:
**Describe the bug**
[CLI] `si hub ls` does not return output in json format, even if  format is set to json in config

1. `si config set log --format json`
2. `si hub ls`

**Steps to Reproduce**
```{
  "configVersion": 1,
  "apiUrl": "http://127.0.0.1:8000/api/v1",
  "debug": false,
  "format": "json",
  "middlewareApiUrl": "https://endpoint/api/v1",
  "env": "production",
  "scope": "",
  "token": "token-abc"
}
```

List hubs:
```
si hub ls |jq
parse error: Invalid numeric literal at line 1, column 6
```
as we can see `jq` can't parse output.

Output without jq:

```
$ si hub ls
Space: org-5cfb82c4-31d4-4f10-93f6-b441858f0396-manager
Hubs [
  {
    id: 'org-5cfb82c4-31d4-4f10-93f6-b441858f0396-host',
    info: {
      created: '2022-05-18T10:31:58.874Z',
      lastConnected: '2022-05-18T10:31:58.876Z'
    },
    healthy: true,
    isConnectionActive: true
  }

```

**Expected behavior**
valid json format from `si hub ls` so e.g `jq` can parse it when we set `format: json` in config

**Version (please complete the following information):**
 - STH version: `si -v 0.22.0`
 - node version: v16.13.2
 - os: Ubuntu 20.04.1

**Additional information like logs, screenshots etc.**

daro1337:
@a-tylenda  same issue with `si space ls` and `si space info`
```bash
$ si space ls
[ { id: 'org-d1767369-373a-439b-9a9e-69ee55562941-manager' } ]

$ si space ls |jq
parse error: Invalid numeric literal at line 1, column 7
```

a-tylenda:
fixed

PR #506, opened 2022-05-19
[CLIfix] `si hub ls` does not return output in json format

a-tylenda:
Issue https://github.com/scramjetorg/transform-hub/issues/505

to verify:
- make sure config is set to beta, and `{ "format":"json" }`
- ts-node packages/cli/src/bin/index.ts space ls |jq
- ts-node packages/cli/src/bin/index.ts hub ls |jq
- ts-node packages/cli/src/bin/index.ts hub info |jq
- ts-node packages/cli/src/bin/index.ts space use <space-name> |jq

### DIFF
--- a/packages/cli/src/lib/commands/config.ts
+++ b/packages/cli/src/lib/commands/config.ts
@@ -1,3 +1,5 @@
+/* eslint-disable no-console */
+
 import { CommandDefinition } from "../../types";
 import { stringToBoolean } from "../../utils/stringToBoolean";
 import { globalConfig, sessionConfig } from "../config";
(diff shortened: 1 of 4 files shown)


daro1337:
I tested it, it works as expected. Thanks:)

daro1337 (approved):

alicja-gruzdz:
great, it works.
Screenshot from 2022-05-19 11-51-30


I'm thinking how to best inform the user that is doesn't work when the `debug` is set to `true`? Maybe to throw an error explaining that? It would be faster than to modify debug to work in json format now.
Screenshot from 2022-05-19 12-10-48

alicja-gruzdz (approved):

daro1337:
@alicja-gruzdz @a-tylenda  after src code check I've found that `si hub use <id>` also uses console.log()
https://github.com/scramjetorg/transform-hub/pull/506/files#diff-304b1fc088d13ee9d93b480e8bdd51bfb23185f84cb1769ddf77b5aacfc63b56R49
https://github.com/scramjetorg/transform-hub/pull/506/files#diff-304b1fc088d13ee9d93b480e8bdd51bfb23185f84cb1769ddf77b5aacfc63b56R58

PR #511, opened 2022-05-20
Fix cucumber tags

jan-warchol:
* Fix BDD tests selection
  
  @ci tag was removed in commit 89671f0b6e421683797c2917345a6c8361b4223b;
  specifying it results in no tests being actually ran.

* Cucumber: remove @ci-python tag, @python is enough

  Since https://github.com/scramjetorg/transform-hub/pull/496, all BDD
  scenarios remaining in the code are enabled on CI, so there is no need
  for an additional @ci-python tag.
  
  Analogous to 89671f0b6e421683797c2917345a6c8361b4223b.
  
* Comment out scenario that should be disabled

  In commit 54c6495d6d5210365b828c63bea0d172b64cd43a this test was
  annotated as buggy. However, since
  https://github.com/scramjetorg/transform-hub/pull/496, all BDD
  scenarios remaining in the code are enabled on CI, so the scenario must
  be commented out to be disabled.


jan-warchol:
I'm not sure why the tests are failing. I thought this was caused by not building the refapps that remained in transform-hub, but it seems they are. I don't know why this doesn't work.

MichalCz:
Continued in #408

PR #516, opened 2022-05-23
Remove unecessary python directory

Tatarinho:


alicja-gruzdz (approved):
btw the testing data files can also be kept in 
https://github.com/scramjetorg/transform-hub/tree/devel/bdd/data/test-data

jan-warchol (approved):

PR #518, opened 2022-05-24
[CLI] Add CLI command `si space info`

a-tylenda:
Click-up [task](https://app.clickup.com/t/24308805/SCP-3787)

There was no possibility to check which space the user is currently in. 
I added `si space info` command which shows space ID and other cpm info:
```
09:31 $ ts-node packages/cli/src/bin/index.ts space info
{ spaceId: 'org-895462b1-e221-48e3-a942-06bba24a6eea-manager' }
{
  service: '@scramjet/manager',
  apiVersion: 'v1',
  version: '0.22.0',
  build: '13d842b'
}
ManagerClient {
  apiBase: 'https://api.beta.scramjet.cloud/api/v1/space/org-895462b1-e221-48e3-a942-06bba24a6eea-manager/api/v1',
  client: ClientUtils {
    apiBase: 'https://api.beta.scramjet.cloud/api/v1/space/org-895462b1-e221-48e3-a942-06bba24a6eea-manager/api/v1',
    fetch: [Function (anonymous)],
    normalizeUrlFn: [Function: normalizeUrl]
  }
}
```

### DIFF
--- a/packages/cli/src/lib/commands/space.ts
+++ b/packages/cli/src/lib/commands/space.ts
@@ -31,6 +31,24 @@ export const space: CommandDefinition = (program) => {
             });
 
     spaceCmd
+        .command("info")
+        /* TODO for future use
+    .argument("[<name>|<id>]")
+    // eslint-disable-next-line max-len
+    .description("display chosen space version if a name|id is not provided it displays a version of a current space")
+    */
+        .description("Display info about the default space")
+        .action(async () => {
+            const spaceId = sessionConfig.getConfig().lastSpaceId;
+            const managerClient = getMiddlewareClient().getManagerClient(spaceId);
+            const version = await managerClient.getVersion();
+
+            displayObject({ spaceId });
+            displayObject(version);
+            displayObject(managerClient);
+        });
+
+    spaceCmd
         .command("list")
         .alias("ls")
         .description("List all existing spaces")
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/commands/space.ts
```diff
@@ -30,6 +30,24 @@ export const space: CommandDefinition = (program) => {
                 throw new Error("Implement me");
             });
 
+    spaceCmd
+        .command("info")
+        /* TODO for future use
```
alicja-gruzdz, line 35: I guess this we're leaving because it is for later implementation, the same as in `host info`?

a-tylenda: We leave for future implementation displaying info of a particular space (by giving its name|id ?? tbc). so you can check the info of any space, not only the one you are currently in. 
Yes, the same with `hub`.

a-tylenda: > I'm only thinking whether not to present the three json configs in an array. This way the result would be a valid json.

Good catch @alicja-gruzdz, thanks! I will put it in an object.

alicja-gruzdz (approved):

alicja-gruzdz:
congrats VALID (RFC 8259) :heavy_check_mark: 
Screenshot from 2022-05-24 12-49-53

patuwwy (approved):

PR #520, opened 2022-05-24
Add a PR template

jan-warchol:
**What?**
Add a pull request template. Also remove old template directory, which works only when a query param is specified in URL.

**Why?**
A template should make it easier to write good PR descriptions. With good PR descriptions:
- reviews should be easier, faster and more thorough
- **keeping up to date** should be easier - we don't have time to read all PRs, but we should be able to read all PR descriptions, and **brief yet informative descriptions should give us just enough insight into what's happening in the project.**
- making release notes will be easier & faster

**Usage:**
- This will be enabled automatically once merged to repo's default branch.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,32 @@
+<!-- If writing isn't your strength, ask @jan-warchol for help ;) -->
+<!-- (or join our Discord https://discord.com/invite/ngXmwvjSYF)  -->
+
+
+**What?**  <!-- Two-sentence summary, understandable for a junior. -->
+
+
+**Why?**  <!-- What is this needed for? You can link to an issue. -->
+
+
+**Usage:**
+<!-- Example (if applicable), how to verify (if not covered by tests). -->
+- 
+- 
+- 
+
+<!--------------------- For non-trivial changes: ---------------------->
+
+**How it works:**
+<!-- Share some starting points for understanding the code. -->
+- 
+- 
+- 
+
+**Gotchas:**
+<!-- Any known shortcomings, alternatives to consider, future todos... -->
+- 
+- 
+- 
+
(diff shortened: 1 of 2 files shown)


jan-warchol:
I suggest that the main goal we should have when writing PR descriptions is to help other team members understand what's going on. In other words, write the descriptions so that anyone from Scramjet will know how a PR affects our project, even if they are not familiar with a particular area of code.

.github/pull_request_template.md
```diff
@@ -0,0 +1,31 @@
+<!-- If writing isn't your strength, ask Janek for help :-) -->
```
MichalCz, line 1: Why change the name of the default.md?

alicja-gruzdz:
Thanks, it will be useful. 
Minor: I'd add `Usage` not only for "For non-trivial changes" but in every situation where we do not cover the change with the tests.

stokowski (approved):
lgtm

alicja-gruzdz (approved):

MichalCz:
Hmm... This sadly breaks down the way the `default.md` worked in `gh` CLI. I reverted the rm.

.github/pull_request_template.md
MichalCz, line 1: This is a public repo - we need to direct people to discord.

jan-warchol: Is this the right link? https://discord.com/invite/ngXmwvjSYF

jan-warchol:
> Hmm... This sadly breaks down the way the `default.md` worked in `gh` CLI. I reverted the rm.

Interesting. I've just checked and having just `.github/pull_request_template.md` works fine for me:
image
(of course this file has to be on a default branch already, that's why I checked in platform-tests repo where this is already merged)

.github/pull_request_template.md
jan-warchol: Added.

PR #521, opened 2022-05-24
Instance limits + AuditStream limit info

patuwwy:
AuditStream Instance Heartbeat:
- is based on Runner monitoring messages here.
- contains memory limit (docker only for now)

```
si seq start - --limits "{ \"memory\": 2048 }" --args="[1000]"
```

docker stats
```
CONTAINER ID   NAME              CPU %     MEM USAGE / LIMIT   MEM %     NET I/O         BLOCK I/O   PIDS
42b48560e970   agitated_edison   0.10%     13.12MiB / 2GiB     0.64%     58.6kB / 76kB   0B / 0B     8
```

AuditStream:
```
{"opState":"ACTIVE","opCode":13012,"objectId":"4fccf1ad-1893-4dca-9888-3005581f88a0","requestorId":"system","receivedAt":1653515019817,"limits":{"memory":2048}}
```


packages/host/src/lib/host.ts
```diff
@@ -721,6 +724,7 @@ export class Host implements IComponent {
 
         csic.on("end", (code) => {
             this.logger.trace("CSIControlled ended", `Exit code: ${code}`);
+            csic.logger.unpipe(this.logger);
```
alicja-gruzdz, line 727: excellent

alicja-gruzdz (approved):
Super useful feature.

Screenshot from 2022-06-01 11-27-06
Screenshot from 2022-06-01 11-27-22

alicja-gruzdz:
I know you have a lot on your plate with the Audit. I was thinking that we could add information about set limit also to Instance API endpoint `/`. If not now, than maybe later on, it could be useful if we want to provide this info in the Panel with Instance info.
Screenshot from 2022-06-01 11-37-32

alicja-gruzdz:
I have noticed another thing that we could later add (sorry :wink:)
1) If the requested limit is larger than available RAM we do not signal it, just set the max RAM (which maybe too much, plus it would be later nice to inform the user somehow)
Screenshot from 2022-06-01 11-47-42


Screenshot from 2022-06-01 11-47-20
(yes, I've got 128 GB star_struck:)


2) We can start Instances of which sum of their set limits is larger than available RAM.
Screenshot from 2022-06-01 11-55-49

Maybe the limits behavior is related to how Docker handles it and it's ok. I'm sorry I didn't have time to check it but I thought it's worth mentioning since I've noticed it.

PR #529, opened 2022-05-30
Preparation for BDD tests unification

Tatarinho:
There is a need for small adjustments in bdd host-steps and utils to provide working tests in next PRs.

### DIFF
--- a/bdd/lib/utils.ts
+++ b/bdd/lib/utils.ts
@@ -194,10 +194,13 @@ export async function waitUntilStreamEquals(stream: Readable, expected: string, 
                 response += chunk;
 
                 if (response === expected) return expected;
+                if (response.length >= expected.length) {
+                    assert.equal(response, expected);
+                }
             }
             throw new Error("End of stream reached");
         })(),
-        defer(timeout).then(() => { throw new Error("Timeout reached"); })
+        defer(timeout).then(() => { assert.equal(response, expected); })
     ]);
 
     return response;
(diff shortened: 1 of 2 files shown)


bdd/lib/utils.ts
```diff
@@ -194,10 +194,13 @@ export async function waitUntilStreamEquals(stream: Readable, expected: string,
                 response += chunk;
 
                 if (response === expected) return expected;
+                if (response.length >= expected.length) {
+                    assert.equal(response, expected);
+                }
```
jan-warchol, line 199: maybe I'm missing something, but isn't this redundant? we have previous check for the situation when lengths are equal, and when they are unequal, the assertion will fail.

jan-warchol (commented):
I don't see anything wrong, but I'm not sure what this change gives us. Could you explain?

Tatarinho:
> I don't see anything wrong, but I'm not sure what this change gives us. Could you explain?

Change in host-steps.ts is related to js sequences with debug args from this pr: https://github.com/scramjetorg/reference-apps/pull/2

Without it we are not able to read output response. As we want to have unified tests for python and js sequences, we need some extra adjustments also in hoststeps etc. to be sure everything works the same for both languages.

bdd/lib/utils.ts
Tatarinho: For now we don't see reason why it failed.

alicja-gruzdz (approved):
As extra we could check if each chunk sub-matches `expected` and break if not

PR #531, opened 2022-05-31
Instance status info

kociolekscramjet:
**What?**
Added instance status to instance getInfo and enhanced instances list with more data.

**Why?**
Instance status is needed to properly handle instances in panel

**Usage:**
Start instance locally and get info about it (instructions in progress)

### DIFF
--- a/packages/host/src/lib/csi-controller.ts
+++ b/packages/host/src/lib/csi-controller.ts
@@ -70,6 +70,7 @@ export class CSIController extends TypedEmitter<Events> {
         started?: Date;
         ended?: Date;
     } = {};
+    status: "initializing" | "starting" | "running" | "finishing" | "ended" | "errored";
     provides?: string;
     requires?: string;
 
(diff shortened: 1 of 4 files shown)


alicja-gruzdz (approved):
I think adding this information will be useful especially in k8s where the Runner works slower. Also, it will be insightful to see the Instance exit status.

I didn't manage to catch all states, but here is `running` and `errored`:
Screenshot from 2022-06-01 12-46-23

Screenshot from 2022-06-01 12-46-55

we get `errored` on `stop` and `kill` but this is how due to how it is currently implemented in Runner + maybe the lack of implementing stopHandler in Sequence (I used the https://github.com/scramjetorg/reference-apps/tree/main/js/csv-transform)
https://github.com/scramjetorg/transform-hub/blob/devel/packages/runner/src/runner.ts#L239
Screenshot from 2022-06-01 13-12-10

PR #532, opened 2022-05-31
Remove cpu-features and systeminformation modules

MichalCz:
**What?**  

Remove dependency on systeminformation.

**Why?**  

Because it was bad.

**Usage:**

If tests pass, it means it worked - as long as loadcheck does work all is fine.


MichalCz:
This currently crashes on request for health. I need to fix that.

patuwwy:
Rebased

alicja-gruzdz (approved):
excellent, thanks @patuwwy

PR #534, opened 2022-06-01
Fix the issue when pod was deleted before k8s client got it status

patuwwy:
**What?**  
Increase time for InstanceAdapter to finish.

**Why?**
K8S runner pod was deleted before Pod status (Failed/Succedded) was obtained what was causing 404 and instance adapter/csic to fail


jan-warchol (approved):

daro1337 (approved):
I attended a debugging session, the solution works :) GJ @patuwwy

PR #539, opened 2022-06-02
Tag docker images with commit, don't overwrite release tag

jan-warchol:
**What?**  
Do not "reuse" docker tags, especially release tags. Tag images with current commit ID.

**Why?**  
Once a "release" tag (like 0.24.1) is used on an image, it should never
be used on other images to avoid confusion and wrong deployments.
Using a commit ID to tag images will make it easy to refer to specific images elsewhere.

**Usage:**

Running `yarn build:docker` e.g. in runner package should result in something like
```
...
Successfully tagged runner:df744aae095832a1ac2c8577ad1c49d3fb434745
```
Indicating it was done on commit df744aae.

PR #540, opened 2022-06-02
Share docker images via repository instead of artifacts

jan-warchol:
**What?**  
Instead of wrapping docker images in an artifact, use our nexus repo for publishing them in build jobs and pulling in test jobs.

**Why?**  
This way docker images can be easily shared in jobs across repositories. E.g. images built here could then be used in CPM CI (potential speed gains).

**Gotchas:**

- this is just an example for pre-runner. It will have to be done for other components once we agree on a direction.
- this requires configuring GHA of this repo with secrets for accessing nexus repo.
- this PR depends on https://github.com/scramjetorg/transform-hub/pull/539

### DIFF
--- a/.github/workflows/build-docker-prerunner.yml
+++ b/.github/workflows/build-docker-prerunner.yml
@@ -8,6 +8,9 @@ on:
         default: '16.x'
         type: string
 
+env:
+  nexus_repo: repo.int.scp.ovh
+
 jobs:
   build-docker-prerunner-image:
     name: Build pre-runner docker image (Nodejs ${{ inputs.node-version }})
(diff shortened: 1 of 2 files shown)


stokowski:
This won't work while we are using public gh runners and nexus is not exposed to the world (and we will not expose it).

jan-warchol:
> This won't work while we are using public gh runners and nexus is not exposed to the world (and we will not expose it).

Ah, sorry, I forgot about the VPN. Right. So, maybe we could use docker hub? We could setup a separate dev repo so that we won't overflow scramjetorg with dev images.

jan-warchol:
BTW we could setup a separate job running on self-hoster runner that would add the image to our private registry (as the runner would be inside VPN)

PR #544, opened 2022-06-03
CLI Configuration using profiles

ErykSol:
This PR changes configuration handling in CLI package.

Most important changes made to configuration:
-There are 3 configuration types inside CLI for different purpose:
	a) Profile configuration- all configurations that can be manipulated and set by user.
	b) Session configuration- configuration used inside of CLI to handle all lastSequenceId, lastInstanceId etc. Configuration is kept alive in terminal (shell session time of live) that uses CLI (as outcome every terminal will have it's own lastXXX params).
	c) Si configuration- global configuration used by all CLI processes.
-Profile configuration can be provided in two ways: 
	a) path to outside configuration file provided by user (read only file, requires all configuration fields in file)
	b) profile name to configuration handled and manipulated by CLI resources filesg
-There is always at least a default profile file.
-Order of precedence for profile configuration is added (If profile config is set with many options only highest precedence is used): 
	1) `si someCliCommand --config-path some/path/toFile`
	2) `si someCliCommand  --config profileName`
	3) process.env: `SI_CONFIG_PATH=some/path/toFile si someCliCommand` 
	4) process.env: `SI_CONFIG=profileName  si someCliCommand`
	5) `si config profile use profileName` (sets default profile to use by all CLI instances)
	6) if no other method is used default profile is used inside CLI
-User calling i.e. `si config set xxx` changes value in currently used profile (default profile aswell), if config is provided from user file exception is thrown.
-Displayed messages are formated with `si config set log --format json/pretty` options
-if error ocured stack trace can be shown/hide with `si config log --debug true/false` (or 1/0, TRUE/FALSE)

Additionoal changes:
-added validation of config fields with js validator package
-errorHandler moved to separate file for readability, and upgraded to use debug and format options from config.
-small readability fixes
-list of profiles is sorted alphabeticallly
-format and debug options in profile config are now nested in log object (i.e. log: {debug: true, format: "json"}) to match CLI api structure
-all logging messages through out the aplication is now handled by displayError, displayMessage methods instead of console.log

### DIFF
--- a/packages/cli/src/lib/config.ts
+++ b/packages/cli/src/lib/config.ts
@@ -198,10 +198,14 @@ export class ProfileConfig extends DefaultFileConfig {
     set readOnly(readOnly: boolean) { this.readOnlyConfig = readOnly; }
     get readOnly() { return this.readOnlyConfig; }
 
-    checkIfWritable() { if (this.readOnlyConfig) throw Error("Unable to write for read only configuration"); }
+    checkIfWritable() { 
+        if (this.readOnlyConfig) throw Error("Unable to write for read only configuration"); 
+    }
+
     setConfig(config: any): boolean {
         this.checkIfWritable();
         if (typeof config !== "object" || !this.validateConfigPart(config)) return false;
+        
         const { log: currentLog, ...currentConfig } = this.getConfig();
         const { log: newLog, ...newConfig } = config;
         const overlap = { ...currentConfig, ...newConfig, log: { ...currentLog, ...newLog } };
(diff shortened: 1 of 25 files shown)


MichalCz (approved):
Very nice code. Fix up the tests and check the eslint errors.

@daro1337 take your 1337 glance at that and verify.

daro1337:
1. config create <profile> creates profile from current profile, that leads to copy also:
```bash
  lastPackagePath: '',
  lastInstanceId: '',
  lastSequenceId: '',
  lastSpaceId: '',
  lastHubId: ''
```
And this values could not exist on new profile. We should create profile from `basic default configuration`

packages/cli/src/lib/config.ts
```diff
@@ -325,7 +325,7 @@ const createDefaultProfileIfNotExist = () => {
     const { profile } = siConfig.getConfig();
 
     if (profile === "" || !profileExists(profile))
-        siConfig.setProfile(defaultConfigProfileFile);
+        siConfig.setProfile("default");
```
daro1337, line 328: Nice, tested - works much more predictably.

Tatarinho (approved):

MichalCz:
obraz

Verified!

PR #547, opened 2022-06-06
Add multilang test for output availablility after completion

jan-warchol:
**What?**  
Replace js-only "E2E-001 TC-002 Test stdio available after the sequence is completed" with new "E2E-015 TC-015 Output is available after the sequence is completed" that supports both py and js.

**Why?**  
We need better coverage for python, to ensure it behaves the same as js.

**Notes:**

- using output vs stdout shouldn't make any difference
- I think there is no need to wait before sending input
- explicitly check that the instance is not running anymore


alicja-gruzdz:
@a-tylenda will do a re-design and re-factor of STH tests and she will take this proposition into account.
We're closing this PR and the test will come with the STH tests re-design.

PR #551, opened 2022-06-07
Instance lifecycle test scenarios

jan-warchol:
**What?**  
Describe how Instance statuses (returned by instance getInfo) and InstanceClient methods related to the instance lifecycle should behave, in the form of BDD test scenarios (as I understood them).

**Why?**  
Communication about instances between backend and frontend code relies on InstanceClient and instance statuses returned by `InstanceClient.getInfo()` (or `si inst info`). These statuses used to be a bit unreliable; this is an attempt to specify how exactly they should behave.

**Gotchas:**

- I know @kociolekscramjet is making some changes to the statuses, I may have missed something here.
- There are only scenarios for now, no step definitions. Will implement them when we agree on the scenarios themselves (true BDD style!)
- I know there are some delays in "finishing" the instance (so that streams can be accessed after computation finishes), this may need clarification.

### DIFF
--- a/bdd/features/instance-lifecycle.feature
+++ b/bdd/features/instance-lifecycle.feature
@@ -1,0 +1,32 @@
+Feature: Instance lifecycle API behavior
+
+    Background:
+        Given host is running
+
+    Scenario: Starting an instance
+        Given sequence "hello" is available
+        When SequenceClient.start is executed
+        Then it should return an InstanceClient
+        And instance status should be "starting"
+        When instance starts producing output
+        Then instance status should be "running"
+
+    Scenario: Instance completes successfully
+        Given instance "hello" is running
+        When instance output is completed
+        Then instance status should immediately change to "finishing"
+        When wait until runner has ended execution
+        Then instance status should immediately change to "completed"
+
+    Scenario Outline: Interrupting an instance
+        Given instance "hello" is running
+        When InstanceClient's <method> is executed
+        Then <method> request should resolve successfully
+        And instance status should immediately change to <status in progress>
+        When wait until runner has ended execution
+        Then instance status should immediately change to <final status>
+
+        Examples:
+            | method | status in progess | final status |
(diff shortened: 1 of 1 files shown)


bdd/features/instance-lifecycle.feature
```diff
@@ -0,0 +1,32 @@
...
+        And instance status should be "starting"
+        When instance starts producing output
+        Then instance status should be "running"
+
+    Scenario: Instance completes successfully
+        Given instance "hello" is running
+        When instance output is completed
+        Then instance status should immediately change to "finishing"
+        When wait until runner has ended execution
+        Then instance status should immediately change to "completed"
```
jan-warchol, line 19: I'm not sure about this. Do we actually need `finishing` status? Is there a real difference between output being completed (i.e. stream is ended and returned) and runner shutdown? How does runner exit delay affect this? I think we should discuss what @kociolekscramjet and @agne6ka expect here and whether backend can behave in such way.

MichalCz:
I think we'd need to obtain these state transitions through audit stream and confirm they are happening in order. This check may end up quite flaky.

bdd/features/instance-lifecycle.feature
patuwwy: Runner has 10seconds exit delay. So when sequence is ended (stream ended/promise resolved, smth returned) and SEQUENCE_COMPLETED control message is sent there is 10secs of `sleep` in Runner before it exits. this is the `finishing` status. when runner returns exitCode we determine if instance is `errored` or `completed`

MichalCz:
@jan-warchol can you check if we can connect to monitoring stream and check the status changes there?

MichalCz (approved):
good stuff.

alicja-gruzdz:
It has been a while :) the proposed BDD scenario will be taken into consideration by @a-tylenda when she will be working on the STH tests re-design, thanks!

PR #561, opened 2022-06-10
Feat/api server op raw body

kociolekscramjet:
**What?** 
Basically it's just the rawBody flag added to the router.op method for endpoints which require raw body data.

**Why?**
Stripe webhook event signature verification is done on raw request body data. Hence the flag is needed to turn off the default JSON.parse 

**Usage:**
It's the last parameter.
```ts
        this.api.op(
            "post",
            "/webhook/payments",
            (req: ParsedMessage) => this.paymentsWebhookHandler(req),
            undefined,
            true
        );
```

**How it works:**
The handlers/op.ts was refactored. To understand it better the core feature here is parameter named rawBody. Scan code in search for that variable.

### DIFF
--- a/packages/api-server/src/handlers/op.ts
+++ b/packages/api-server/src/handlers/op.ts
@@ -130,19 +130,19 @@ export function createOperationHandler(router: SequentialCeroRouter): APIRoute["
      * @param {IncomingMessage} req Request object.
      * @param {ServerResponse} res Server response.
      * @param {ControlMessageCode} message Control message.
-     * @param {ICommunicationHandler} conn Communication handler.
+     * @param {ICommunicationHandler} comm Communication handler.
      * @returns void
      */
     const opControlMessageHandler = async <T extends ControlMessageCode>(
         req: ParsedMessage,
         res: ServerResponse,
         message: T,
-        conn: ICommunicationHandler
+        comm: ICommunicationHandler
     ) => {
         // eslint-disable-next-line no-extra-parens
         const obj = ((await getData(req)) as Array<any>)[1] as MessageDataType<T>;
 
-        await conn.sendControlMessage(message, checkMessage(message, obj));
+        await comm.sendControlMessage(message, checkMessage(message, obj));
 
         res.writeHead(StatusCodes.ACCEPTED, ReasonPhrases.ACCEPTED, { "content-type": "application/json" });
         return res.end(JSON.stringify({ accepted: true }));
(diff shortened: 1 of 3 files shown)


packages/api-server/src/handlers/op.ts
```diff
@@ -17,16 +25,15 @@ const logger = new ObjLogger("API");
...
+     * Checking content-type and getting encoding from request
      * @param {IncomingMessage} req Request object.
-     * @returns JSON object.
+     * @param {boolean} rawBody Flag if the body will be parsed.
+     * @returns BufferEncoding string
      */
-    const getData = async (req: IncomingMessage): Promise<object|undefined> => {
-        if (!req.headers["content-type"])
-            throw new CeroError("ERR_INVALID_CONTENT_TYPE");
+    const getEncoding = (req: IncomingMessage, rawBody: boolean = false): BufferEncoding => {
```
MichalCz, line 33: I'd rather we did this like this:

```suggestion
    const getEncoding = (req: IncomingMessage, { rawBody: boolean = false }): BufferEncoding => {
```

WDYT?

packages/api-server/src/handlers/op.ts
```diff
@@ -48,17 +65,89 @@ export function createOperationHandler(router: SequentialCeroRouter): APIRoute["
...
+        const body = await getBody(req, encoding);
+
         try {
-            if (!out) {
+            if (!body) {
                 return undefined;
             }
 
-            return JSON.parse(out);
+            return rawBody ? body : JSON.parse(body);
```
MichalCz, line 87:
```suggestion
            return body && rawBody ? body : JSON.parse(body);
```

packages/api-server/src/handlers/op.ts
```diff
@@ -48,17 +65,89 @@ export function createOperationHandler(router: SequentialCeroRouter): APIRoute["
...
+     */
+    const getData = async (req: IncomingMessage, rawBody?: boolean): Promise<object | undefined> => {
+        const encoding = getEncoding(req, rawBody);
+        const body = await getBody(req, encoding);
+
         try {
-            if (!out) {
+            if (!body) {
                 return undefined;
             }
```
MichalCz, line 85:
```suggestion
```

packages/api-server/src/handlers/op.ts
```diff
@@ -48,17 +65,89 @@ export function createOperationHandler(router: SequentialCeroRouter): APIRoute["
...
+    /**
+     * Tries to parse data from the request body.
+     *
+     * @param {IncomingMessage} req Request object.
+     * @param {boolean} rawBody Flag if the body will be parsed.
+     * @returns JSON object.
+     */
+    const getData = async (req: IncomingMessage, rawBody?: boolean): Promise<object | undefined> => {
+        const encoding = getEncoding(req, rawBody);
+        const body = await getBody(req, encoding);
```
MichalCz, line 80: This potentially throws so should be in the try/catch.

MichalCz (commented):
Take a look at the suggestions.

MichalCz (approved):
Please move the `getBody` call to within the try/catch.

MichalCz (approved):

PR #563, opened 2022-06-14
Disable Nagle algorithm

patuwwy:
**What?**  
Disable Nagle algorithm

**Why?**  
checking performance

### DIFF
--- a/packages/verser/src/lib/verser.ts
+++ b/packages/verser/src/lib/verser.ts
@@ -3,6 +3,7 @@ import { Server } from "http";
 import { VerserConnection } from "./verser-connection";
 import { TypedEmitter } from "@scramjet/utility";
 import { ObjLogger } from "@scramjet/obj-logger";
+import { Socket } from "net";
 
 type Events = {
     connect: (connection: VerserConnection) => void;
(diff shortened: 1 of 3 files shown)


MichalCz:
Patryk, can we merge this?

alicja-gruzdz (approved):

patuwwy:
@MichalCz now we can

MichalCz:
I'd like to run this before. @alicja-gruzdz @daro1337 perhaps we should do a patch release before releasing this? So 0.28.4 before this in 0.28.5

ErykSol (approved):

packages/verser/src/lib/verser.ts
```diff
@@ -23,9 +24,12 @@ export class Verser extends TypedEmitter<Events> {
         super();
         this.server = server;
 
-        this.server.on("connect", (req, socket) => {
+        this.server.on("connect", (req, socket: Socket) => {
             this.logger.info("New connection:", req.url);
 
+            socket.setTimeout(0);
```
ErykSol, line 30: As I understand timeout=0 is just for testing, not as permanent solution?

patuwwy: permanent, https://nodejs.org/api/net.html#socketsettimeouttimeout-callback

ErykSol: Ah now I see, thx for claryfying :)

alicja-gruzdz (approved):

patuwwy:
can we merge this one?

alicja-gruzdz:
> can we merge this one?

In my opinion yes since it showed green in CPM.

Issue #537, opened 2022-06-01
LoadCheck doesn't use Config/limits when determining if instance can be run


patuwwy:
LoadCheck uses https://github.com/scramjetorg/transform-hub/blob/devel/packages/sth-config/src/config-service.ts#L37-L42 but doesn't read instance requirements.

> yup. load-check has hardcoded values(mem free) required to run an instance and doesn't include limits from config / limits

_Originally posted by @patuwwy in https://github.com/scramjetorg/transform-hub/issues/521#issuecomment-1143704676_

Tatarinho:
@MichalCz is it finished?

PR #570, opened 2022-06-20
Option for changing memory reserve.

MichalCz:
**What?**  

Add an option in the config that allows changing the minimum memory needed for STH to start more sequences.

**Why?**  

Currently STH doesn't allow starting sequences on RPi zero, as the default limit is 512 megs
and that's how much the RPi Zero has.


MichalCz:
Fixes #537

packages/sth/src/bin/hub.ts
```diff
@@ -27,6 +27,7 @@ const options: STHCommandOptions = program
     .option("--runner-max-mem <mb>", "Maximum mem used by runner")
     .option("--prerunner-image <image name>", "Image used by prerunner")
     .option("--prerunner-max-mem <mb>", "Maximum mem used by prerunner")
+    .option("--safe-operation-limit <mb>", "NUmber of MB reserved by the host for safe operation")
```
patuwwy, line 30:
```suggestion
    .option("--safe-operation-limit <mb>", "Number of MB reserved by the host for safe operation")
```

patuwwy (approved):

a-tylenda (approved):

PR #577, opened 2022-06-24
Python runner support for keepalive

Tatarinho:


packages/python-runner/runner.py
```diff
@@ -146,15 +146,15 @@ async def connect_control_stream(self):
...
-        # canCallKeepAlive = data.get('canCallKeppalive')
         self.logger.info(f'Gracefully shutting down...{data}')
         try:
             timeout = data.get('timeout') / 1000
             await asyncio.sleep(timeout)
+            
+            can_keep_alive = data.get('canCallKeepalive')
+            if can_keep_alive:
+                for handler in self.stop_handlers:
+                    await handler()
```
MichalCz, line 157: This is inconsistent with JS - the `handler()` needs to be called with arguments and regardless of keepAlive. See here [the js implementation](https://github.com/scramjetorg/transform-hub/blob/09859fd88583e8663222be97c63e1679c020ed66/packages/runner/src/runner.ts#L254).

MichalCz (changes requested):

MichalCz (approved):

MichalCz (approved):

Issue #608, opened 2022-07-22
[STH]  Implicit conversion of input when using application/octet-stream


ErykSol:
While using CLI command `si instance input <id> <file> -t application/octet-stream` STH passes <file> to the user sequence as a Buffer implicitly converted to 'utf-8' and not as untouched binary.

To reproduce

1.  download hexdump sequence and `cd` to it:
 https://github.com/scramjetorg/platform-samples/pull/6
3.  run script: `yarn run packTestFiles` (it produces testFiles/testTar.tar.gz file)
4. `yarn build`
In I terminal:
5. `si seq deploy dist/`
6. `si inst log <id_of_inst>`
In II terminal
7. `si inst input <id_of_inst>  testFiles/testTar.tar.gz -t  application/octet-stream`
In III terminal
8. `si inst output <id_of_inst> `

In logs you can check data given on input on sequence:

example of current output:
`{"level":"DEBUG","msg":"--> Output data: \u001f�\b\u0000\u0000\u0000\u0000\u0000\u0000\u0003��M\n�0\u0014F�,%+h���;p�@7 \u001a1X-4\u0011\\�m\u0015Q�t\u0012��=��\u000e\u0002/����B\\�څ<4g�y�DB�cmՏza���,��\nk�֪�Bicm)�J��1�\u0010���
�v��\u001f�7u>S��2�[��\u0012����(\u000f�\u0000�U���U<�6q\u0013�����U�k�!��0���\u00014�\u000e\u001d\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000`\u001e�y�5�\u0000(\u0000\u0000","ts":1658500949654,"from":"Sequence","Runner":{"id":"adf10eb1-731a-4b50-870f-973a7de3ddcf"}}`

On output we should receive hexdump of input data, instead we receive hexdump of input data conferted to (probably) utf-8. Size and format doesn't match.
Fix should provide input receiving same byte data (in chunks) as `cat ./testFiles/testTar.tar.gz | hexdump` and the output of sequence should match hexdump of input file.

Tatarinho:
Python sequence has similar issue.
![image](https://user-images.githubusercontent.com/23729934/180984453-01f5cdec-8e1b-443f-8def-063b90e8558f.png)

Tatarinho:
[...]

PR #612, opened 2022-07-28
Fix handling binary input

patuwwy:
**What?**  
change default input encoding to `binary`

**Why?**  
when passing binary data to `/input` data was encoded to `utf-8` 

see #608


MichalCz (approved):

Tatarinho:
I've commented on [issue](https://github.com/scramjetorg/transform-hub/issues/608) that everything looks ok.

Tatarinho (approved):

Issue #607, opened 2022-07-20
[API] List Topics response is to big


agne6ka:
## **Describe the bug**

List Topics response has to much information.

## **Steps to Reproduce**
1. Create Topic via POST `https://api.beta.scramjet.cloud/api/v1/space/:id/api/v1/sth/:id/api/v1/topic/topicname`
2. List Topics `https://api.beta.scramjet.cloud/api/v1/space/:id/api/v1/sth/:id/api/v1/topics`

```json
[
	{
		"contentType": "application/x-ndjson",
		"stream": {
			"_readableState": {
				"objectMode": false,
				"highWaterMark": 16384,
				"buffer": {
					"head": {
						"data": {
							"type": "Buffer",
							"data": [
								123,
								10,
								32,
								32,
								32,
								32,
								34,
								97,
								/...
							]
						},
						"next": {
							"data": {
								"type": "Buffer",
								"data": [
									10
								]
							},
							"next": null
						}
					},
					"tail": {
						"data": {
							"type": "Buffer",
							"data": [
								10
							]
						},
						"next": null
					},
					"length": 2
				},
				"length": 51,
				"pipes": [],
				"flowing": null,
				"ended": false,
				"endEmitted": false,
				"reading": true,
				"constructed": true,
				"sync": false,
				"needReadable": true,
				"emittedReadable": false,
				"readableListening": false,
				"resumeScheduled": false,
				"errorEmitted": false,
				"emitClose": true,
				"autoDestroy": true,
				"destroyed": false,
				"errored": null,
				"closed": false,
				"closeEmitted": false,
				"defaultEncoding": "utf8",
				"awaitDrainWriters": null,
				"multiAwaitDrain": false,
				"readingMore": false,
				"dataEmitted": false,
				"decoder": null,
				"encoding": null
			},
			"_events": {},
			"_eventsCount": 1,
			"_writableState": {
				"objectMode": false,
				"highWaterMark": 16384,
				"finalCalled": false,
				"needDrain": false,
				"ending": false,
				"ended": false,
				"finished": false,
				"destroyed": false,
				"decodeStrings": true,
				"defaultEncoding": "utf8",
				"length": 0,
				"writing": false,
				"corked": 0,
				"sync": false,
				"bufferProcessing": false,
				"writecb": null,
				"writelen": 0,
				"afterWriteTickInfo": null,
				"buffered": [],
				"bufferedIndex": 0,
				"allBuffers": true,
				"allNoop": true,
[...]

PR #615, opened 2022-07-29
Remove stream object from topics response

patuwwy:
**What?**  
Remove unnecessary data from `/topics` response

**Why?**  
#607 

**Usage:**

```
si topic send mytopic myfile
```
```
si topic ls
```

Response will be similar to:
```
[ { contentType: 'text/plain', localProvider: 'api', topic: 'mytopic' } ]
```

without stream object mentioned in #607


MichalCz (approved):

BKscramjet:
Task linked: [SHN-574 List Topics response is to big](https://app.clickup.com/t/24308805/SHN-574)

PR #621, opened 2022-08-04
Fix delete Sequence with running Instances

Tatarinho:


MichalCz (approved):

MichalCz:
obraz

Verified - still in finalizing. :)

Issue #605, opened 2022-07-20
[API] Stop Sequence response status `500`


agne6ka:
## **Describe the bug**

Stop Instance response status `500`.

## **Steps to Reproduce**
1. There should be a running Instance of Sequence on Hub.
2. Stop the Instance by making call to API `https://api.beta.scramjet.cloud/api/v1/space/:id/api/v1/sth/:id/api/v1/instance/:id/_stop`

## **Expected behavior**

It should respond `202 - Accepted` if the req was accepted and the Instance has status `stopping`.

## **Version (please complete the following information):**
 - Call to API Beta Production
 
 ```json
 {
	"service": "@scramjet/manager",
	"apiVersion": "v1",
	"version": "0.22.0",
	"build": "8b39d26"
}
```
 
 ```json
 {
	"service": "@scramjet/host",
	"apiVersion": "v1",
	"version": "0.24.2",
	"build": "source"
}
```

## **Additional information like logs, screenshots etc.**

- Stop Instance `/instance/:id/_stop` error `500`

![image](https://user-images.githubusercontent.com/12500823/179962777-2a943f50-9924-49bb-bf30-d413fbb06e4c.png)

- Instance list `/instances` desired instance status `stopping` 

![image](https://user-images.githubusercontent.com/12500823/179962733-48c68520-403b-4af1-9272-d54ed06eccba.png)

Tatarinho:
@agne6ka I couldn't reproduce that with js and python sequences

patuwwy:
`_stop` endpoint expects payload with following format

```
"[4001, { timeout: 1000, canCallKeepalive: true }]"
```

PR #624, opened 2022-08-05
Return CONFLICT on `stop` when instance is not running

patuwwy:
**What?**  
Return CONFLICT on `stop` when instance is not running

**Why?**  
to avoid issues like https://github.com/scramjetorg/transform-hub/issues/605

**Review checks:**

- [ ] All STH tests pass


patuwwy:
Doesn't make sense, it has been designed to handle multiple stop commands: https://github.com/scramjetorg/transform-hub/blob/fa39a2976fde8d7f8458c2b724b0a50dbadd508f/bdd/features/e2e/E2E-002-stop.feature#L10

Issue #426, opened 2022-04-12
Topic name validation: topic names including slashes


alicja-gruzdz:
It would be useful to have a validation of topic names:
![Zrzut ekranu z 2022-04-12 13-02-24](https://user-images.githubusercontent.com/8177865/162946539-c1a14a64-4725-4bbd-bb36-8a833a5525e5.png)

alicja-gruzdz:
Done and released. https://github.com/scramjetorg/transform-hub/pull/625

PR #625, opened 2022-08-06
Validate topic name

patuwwy:
**What?**  
Validate topic name when using `si topic send/get`

**Why?**  
https://github.com/scramjetorg/transform-hub/issues/426

**Review checks:**

- [ ] All STH tests pass

### DIFF
--- a/packages/cli/src/lib/commands/topic.ts
+++ b/packages/cli/src/lib/commands/topic.ts
@@ -1,8 +1,17 @@
+/* eslint-disable no-console */
 import { CommandDefinition } from "../../types";
 import { isDevelopment } from "../../utils/envs";
 import { getHostClient, getReadStreamFromFile } from "../common";
 import { profileConfig } from "../config";
 import { displayEntity, displayStream } from "../output";
+
+const validateTopicName = (topicName: string) => {
+    if (topicName.match(/^[a-zA-Z0-9_-]*$/)) {
+        return;
+    }
+
+    throw new Error("Invalid topic name");
+};
 
 /**
  * Initializes `topic` command.
@@ -16,7 +25,7 @@ export const topic: CommandDefinition = (program) => {
         .usage("[command] [options...]")
         .description("Manage data flow through topics operations");
 
-    if (isDevelopment())
+    if (isDevelopment()) {
         topicCmd
             .command("create")
             .argument("<topic-name>")
@@ -25,6 +34,7 @@ export const topic: CommandDefinition = (program) => {
             // FIXME: implement me
                 throw new Error("Implement me");
             });
+    }
 
     topicCmd
         .command("get")
@@ -34,9 +44,10 @@ export const topic: CommandDefinition = (program) => {
             "Specifies data type of <topic-name> (default: application/x-ndjson)"
         )
         .description("Get data from topic")
+        .hook("preAction", (command) => { validateTopicName(command.args[0]); })
         .action(async (topicName) => displayStream(getHostClient().getNamedData(topicName)));
 
-    if (isDevelopment())
+    if (isDevelopment()) {
         topicCmd
             .command("delete")
             .alias("rm")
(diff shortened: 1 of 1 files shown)


BKscramjet:
Task linked: [SHN-583 Topic name validation: topic names including slashes](https://app.clickup.com/t/24308805/SHN-583)

a-tylenda:
Error handling works just fine. I was just wondering about a more precise message regarding the characters that the user should avoid in the name. What do you think?

image
image

packages/cli/src/lib/commands/topic.ts
```diff
@@ -1,9 +1,18 @@
...
 import { getHostClient, getReadStreamFromFile } from "../common";
 import { profileConfig } from "../config";
 import { displayEntity, displayStream } from "../output";
 
+const validateTopicName = (topicName: string) => {
+    if (topicName.match(/^[a-zA-Z0-9_-]*$/)) {
+        return;
+    }
+
+    throw new Error("Invalid topic name");
```
a-tylenda, line 13: Some characters are acceptable and some are not, for example, slash throws an error but backslash is ok:

image

Should we be more precise about the characters that cannot be used in the topic name? Maybe something like this:

```suggestion
    throw new Error("Invalid topic name! Please do not use special characters in the topic name, such as: / ! @ # % ^ & * ( ) + | < >");
```

a-tylenda (approved):

packages/cli/src/lib/commands/topic.ts
patuwwy: actually `si topic send topic\name READM.md` creates `topicname` topic due to escape character `\`

patuwwy:
but yes,
```
curl --url 'http://0.0.0.0:8000/api/v1/topic/topic\name' -d "data"
```
```
si topic ls
[
  {
    contentType: 'application/x-www-form-urlencoded',
    localProvider: 'api',
    topic: 'topic\\name'
  }
]
```

packages/cli/src/lib/commands/topic.ts
MichalCz: I don't actually see anything wrong with `/` if properly escaped.

packages/cli/src/lib/commands/topic.ts
```diff
@@ -1,9 +1,18 @@
+/* eslint-disable no-console */
 import { CommandDefinition } from "../../types";
 import { isDevelopment } from "../../utils/envs";
 import { getHostClient, getReadStreamFromFile } from "../common";
 import { profileConfig } from "../config";
 import { displayEntity, displayStream } from "../output";
 
+const validateTopicName = (topicName: string) => {
+    if (topicName.match(/^[a-zA-Z0-9_-]*$/)) {
```
MichalCz, line 9:
```suggestion
    if (topicName.match(/^[a-zA-Z0-9_-+,\.]*$/)) {
```

MichalCz: Maybe the `+` as that's commonly converted to space in `url`s...

alicja-gruzdz:
> but yes,
> 
> ```
> curl --url 'http://0.0.0.0:8000/api/v1/topic/topic\name' -d "data"
> ```
> 
> ```
> si topic ls
> [
>   {
>     contentType: 'application/x-www-form-urlencoded',
>     localProvider: 'api',
>     topic: 'topic\\name'
>   }
> ]
> ```
> 
> we also need to validate topic names when creating topics on host

I added https://app.clickup.com/t/24308805/SHN-816

packages/cli/src/lib/commands/topic.ts
alicja-gruzdz: I'd stay with unreserved and safe url chars, also due to user friendly looking urls.

a-tylenda (approved):

PR #628, opened 2022-08-07
instance/inout endpoint

patuwwy:
**What?**  
```
POST    INSTANCE_BASE/inout
```

Adds `POST` instance's `input` endpoint which provides instance's `output` in response

**Why?**  
To trasform data with one request

**Usage:**
example. instance transforms ```inputpayload => "Echo: " + inputpayload```. 
`POST` sequence source to instance input and get response in single request.
```
curl http://localhost:8000/api/v1/instance/ef69602d-f494-429d-8fa6-77832d47c5cd/inout --no-buffer -H "Content-Type: text/plain" -d @dist/index.js
Echo: "use strict";Object.defineProperty(exports, "__esModule", { value: true });const stream_1 = require("stream");const mod = [    function (input) {        const out = new stream_1.PassThrough();        this.logger.trace("Instance started");        input            .map((data) => "Echo: " + data + "\n")            .pipe(out);        return out;    }];exports.default = mod;//# sourceMappingURL=index.js.map
```

**How it works:**
Still needs a work. it is a POC.

**Review checks:**

- [x] All STH tests pass

### DIFF
--- a/docs/adapters/classes/KubernetesInstanceAdapter.md
+++ b/docs/adapters/classes/KubernetesInstanceAdapter.md
@@ -50,7 +50,7 @@ Adapter for running Instance by Runner executed in separate process.
 
 #### Defined in
 
-[kubernetes-instance-adapter.ts:34](https://github.com/scramjetorg/transform-hub/blob/HEAD/packages/adapters/src/kubernetes-instance-adapter.ts#L34)
+[kubernetes-instance-adapter.ts:35](https://github.com/scramjetorg/transform-hub/blob/HEAD/packages/adapters/src/kubernetes-instance-adapter.ts#L35)
 
 ___
 
@@ -60,7 +60,7 @@ ___
 
 #### Defined in
 
-[kubernetes-instance-adapter.ts:37](https://github.com/scramjetorg/transform-hub/blob/HEAD/packages/adapters/src/kubernetes-instance-adapter.ts#L37)
+[kubernetes-instance-adapter.ts:38](https://github.com/scramjetorg/transform-hub/blob/HEAD/packages/adapters/src/kubernetes-instance-adapter.ts#L38)
 
 ___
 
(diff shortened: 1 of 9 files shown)


BKscramjet:
Task linked: [SHN-647 Could input and output streams be Duplexes?](https://app.clickup.com/t/24308805/SHN-647)

MichalCz (approved):

MichalCz:
Looks good, tests, docs, verify and the lot needed. :)

a-tylenda:
Verified:

I created a test seq for JS with index.js:

```js
const Stream1 = require("stream");

const mod = [
    function(input) {
        const out = new Stream1.PassThrough();

        this.logger.trace("Instance started");
        input
            .map((data) => "Echo: " + data + "\n")
            .pipe(out);
        return out;
}];

exports.default = mod;
```

I added package.json and send seq to the local hub, started seq and sent a curl to `/inout` endpoint:

```bah
curl -X POST --include http://localhost:8000/api/v1/instance/673a5e65-dc97-4ae0-b448-7410940fa3ea/inout \
--header "Content-Type: text/plain" \
--header "Transfer-Encoding: chunked" \
--no-buffer -T -
```

Then I typed in some data and as a result I got:
```bash
✔ ~/shared/transform-hub/bdd/data/sequences [feature/inout|…1⚑ 10] 
10:00 $ curl -X POST --include http://localhost:8000/api/v1/instance/673a5e65-dc97-4ae0-b448-7410940fa3ea/inout \
> --header "Content-Type: text/plain" \
> --header "Transfer-Encoding: chunked" \
> --no-buffer -T -
HTTP/1.1 100 Continue

HTTP/1.1 100 Continue

Sending sample data to inout endpoint

HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Date: Tue, 11 Oct 2022 10:01:06 GMT
Connection: keep-alive
Keep-Alive: timeout=5
Transfer-Encoding: chunked

Echo: Sending sample data to inout endpoint
```

I also checked it out for python sequence and `--runtime-adapter=process`

```py
from scramjet.streams import Stream

def run(context, input: Stream):
    return input.map(lambda data: f"Echo: {data}")
 ```

Works!

a-tylenda (approved):

alicja-gruzdz (approved):

PR #631, opened 2022-08-09
Fix si seq send does not return valid exit codes

Tatarinho:
Steps to Reproduce:

```
yarn start:dev:cli seq send simple-counter-js.tar.gz; echo $?
yarn run v1.22.19
$ ts-node packages/cli/src/bin/index.ts seq send simple-counter-js.tar.gz
Error: File "simple-counter-js.tar.gz" not found.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
1
```

### DIFF
--- a/packages/cli/src/lib/common.ts
+++ b/packages/cli/src/lib/common.ts
@@ -119,7 +119,8 @@ export const packAction = async (directory: string, { output }: { output: Writab
     // TODO: error handling?
     // TODO: check package contents?
     await access(packageLocation, F_OK | R_OK).catch(() => {
-        throw new Error(`${packageLocation} not found.`);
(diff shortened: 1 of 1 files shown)


patuwwy:
IMO we should introduce actionSafeHandler method and use it for all `si` commands

daro1337 (changes requested):
This only handle `file not found error` what with other errors?
 
```
Expected behavior
Any error from CLI should return exit code != 0
```

patuwwy (changes requested):
should catch all exceptions

patuwwy:
Replaced with #665

PR #636, opened 2022-08-10
Add missing commands to the start snippet

patuwwy:
**Review checks:**

- [x] All STH tests pass

### DIFF
--- a/packages/sth/README.md
+++ b/packages/sth/README.md
@@ -30,7 +30,19 @@ scramjet-transform-hub
 Depending on your machine this may take some time. When it's done the Hub should be running and you should see initial logs showing that the API server has been started on port 8000, something like this:
 
 ```shell
-2021-07-07T18:19:36.808Z info (object:Host) API listening on port: localhost:8000
+$ scramjet-transform-hub
+2022-08-18T07:55:13.135Z INFO  Host Log Level [ 'TRACE' ]
+2022-08-18T07:55:13.137Z TRACE Host Host main called [ { version: '0.27.0' } ]
+2022-08-18T07:55:13.155Z INFO  Host Will use the "docker" adapter for running Sequences
+2022-08-18T07:55:13.157Z INFO  SocketServer SocketServer on [ { address: '::', family: 'IPv6', port: 8001 } ]
+2022-08-18T07:55:13.159Z INFO  Host API on [ '0.0.0.0:8000' ]
+2022-08-18T07:55:13.160Z INFO  Host You don't need to maintain your own server anymore [
+  {
+    'Check out': 'Scramjet Cloud Platform',
+    here: 'https://scr.je/join-beta-sth'
+  }
+]
+
 ```
 
 Now create an application, let's say you want to get the currency rates every 10 seconds and do something. In a clean folder save this as `index.js`:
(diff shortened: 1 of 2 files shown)


patuwwy:
this snippet still needs rework. Why?
```
$ SEQ_ID=$(si seq send ./package.tar.gz)
Request ok: http://127.0.0.1:8000/api/v1/sequence status: 200 OK
$ echo $SEQ_ID
SequenceClient { _id: 'c9ab1abf-5a5e-49b3-9a89-24c35fab06ec', host: HostClient { apiBase: 'http://127.0.0.1:8000/api/v1', client: ClientUtils { apiBase: 'http://127.0.0.1:8000/api/v1', fetch: [Function (anonymous)], normalizeUrlFn: [Function: normalizeUrl], log: [Object] } }, sequenceURL: 'sequence/c9ab1abf-5a5e-49b3-9a89-24c35fab06ec' }
```

default format for cli is pretty https://github.com/scramjetorg/transform-hub/blob/devel/packages/cli/src/lib/config.ts#L149

patuwwy:
Apart from that we can show `si seq deploy` version

packages/sth/README.md
```diff
@@ -77,7 +77,11 @@ Copy a content below and save it as `package.json` file:
 Open a terminal run your program on the hub:
 
 ```bash
-si pack /path/to/my/folder -o ~/package.tar.gz # compress the app to a package
+cd /path/to/my/folder
+
+npm install                                    # install dependencies
+
+si seq pack . -o ~/package.tar.gz              # compress the app to a package
```
a-tylenda, line 84: These changes must be implemented in `readme.mtpl` file → packages/sth/src/readme.mtpl

a-tylenda (changes requested):
Changes should be added in `readme.mtpl` instead of `readme.md`

alicja-gruzdz:
That's a typical mistake I always make with mtpl :) 
Is there any help still needed here?

patuwwy:
yes, should we `si config set log --format json` to use 

SEQ_ID=$(si seq send ./package.tar.gz | jq .id)

or we can just explain `-`

and do

```
si seq send ./package.tar.gz
si seq start - --args [$APIKEY, "BTC", "EUR"]
si inst stdout -
```
btw what is package.tar.gz ? why do i need api key before first steps?

a-tylenda:
> > yes, should we `si config set log --format json` to use
> > ```
> > SEQ_ID=$(si seq send ./package.tar.gz | jq .id)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or we can just explain `-`
> 
> Thanks for explaining. So I see it as a question of what will be easier to use. Also, if we use consistently in the documentations and samples assigning the result of command to a variable then I would change the default CLI to json (omit the additional step to set the formatting to json). Or modify the documentation everywhere accordingly. I think json is such a standard format for developers that nobody will be scared of it when they see it as default. But maybe I underestimate the --pretty cuteness factor :) @a-tylenda what do you think?

I would stick to CLI. I will propose some changes in a commit ✌🏼

alicja-gruzdz:
> I would stick to CLI. I will propose some changes in a commit ✌🏼

excellent, thanks!

a-tylenda (approved):

alicja-gruzdz (approved):

packages/sth/README.md
```diff
@@ -74,17 +87,41 @@ Copy a content below and save it as `package.json` file:
...
+
+# alternatively - param will send the last packaged sequence
+si seq start - --args [\"BTC\",\"EUR\"]
+# <instance-id> will be returned
+
+# see the output from the program by reading the stdout stream
+si inst stdout <instance-id>
+
+# alternatively use - param instead of <instance-id>
+si inst stdout -
```
alicja-gruzdz, line 117: nicely explained

PR #647, opened 2022-08-16
sth add error msg on invalid instace id

ErykSol:
System should inform that instance does not exist while requesting instance details. Currently we response only with statusCode 404, this PR adds message in response body:
{  "error": "The instance <id> does not exist."}

### DIFF
--- a/packages/host/src/lib/host.ts
+++ b/packages/host/src/lib/host.ts
@@ -423,7 +423,8 @@ export class Host implements IComponent {
 
         if (!sequenceInfo) {
             return {
-                opStatus: ReasonPhrases.NOT_FOUND
+                opStatus: ReasonPhrases.NOT_FOUND,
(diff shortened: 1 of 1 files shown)


patuwwy (approved):

patuwwy:
But why can't you assume that instance doesn't exist when status is 404?

MichalCz (approved):

patuwwy (approved):

alicja-gruzdz:
> But why can't you assume that instance doesn't exist when status is 404?

true. We talked during the front planning about the need to provide the common error messages and this is the step in this direction. There is more error messages design coming and I know @ErykSol will involve us in the discussion later on.

alicja-gruzdz:
Just to keep the tradition of verify (I did on devel since it's merged)
Screenshot from 2022-08-17 18-15-31

PR #653, opened 2022-08-18
[STH] Add missing repository field to /sequences

ErykSol:
This PR adds missing "repository" field from parsed sequence package.json to {apiBase}/sequences endpoint


alicja-gruzdz (approved):
Lookin' good :man_dancing: 
Screenshot from 2022-08-18 18-02-32

I only see we will need to correct display of objects in the CLI.

patuwwy:
can we use `console.dir(obj, { depth: null })` to show deeper object's content ?

PR #654, opened 2022-08-19
[CLI] display object depth null

ErykSol:
This PR sets display of object depth without limit.


alicja-gruzdz (approved):
I verified comparing the old display of repository information with the new one.
Old:
Screenshot from 2022-08-18 18-02-32

New:
Screenshot from 2022-08-19 09-15-27

Issue #592, opened 2022-07-07
[CLI] si seq send does not return valid exit codes


daro1337:
**Describe the bug**
CLI does not return valid exit codes

**Steps to Reproduce**
```bash
$ si seq send simple-counter-js.tar.gz; echo $?
Error: Failed to request 'http://127.0.0.1:8000/api/v1/sequence' with code ECONNREFUSED.
0
```

```bash
$ si seq send simple-counter-js.tar.gz1; echo $?
Error: File "simple-counter-js.tar.gz1" not found.
0
```
Would be nice to check other `si` commands for this issue

**Expected behavior**
Any error from CLI should return exit code != 0

**Version (please complete the following information):**
 - STH version: 0.25.x
 - node version: 16.x
 - os: Ubuntu
 - CLI version: 0.25.0

**Additional information like logs, screenshots etc.**

PR #665, opened 2022-09-01
[CLI] fix returned exit codes

ErykSol:
This PR provieds fix for https://github.com/scramjetorg/transform-hub/issues/592

Some CLI commands although returning in logs error messages, still returns exit status value as 0. This PR fixes it by moving handling of errors into global error handler.

- [x] Verify and confirm operation (please post a screenshot) 
- [x] All STH tests pass

Example result:
image

### DIFF
--- a/bdd/features/e2e/E2E-010-cli.feature
+++ b/bdd/features/e2e/E2E-010-cli.feature
@@ -1,6 +1,6 @@
 Feature: CLI tests
 
-This feature checks CLI functionalities without topics
+    This feature checks CLI functionalities without topics
 
     @ci-api @cli
     Scenario: E2E-010 TC-001 Test 'si --help' and 'si --version' display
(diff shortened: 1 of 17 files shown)


ErykSol:
> does it replace #631 ?

Yes

MichalCz (approved):

alicja-gruzdz (approved):
Minor thins, but I've noticed these two:
1. "Couldn't find scope" printed twice
I believe from these two places:
https://github.com/scramjetorg/transform-hub/blob/fix/cli_valid_exit_codes/packages/cli/src/lib/helpers/scope.ts#L11

https://github.com/scramjetorg/transform-hub/blob/fix/cli_valid_exit_codes/packages/cli/src/lib/commands/scope.ts#L35

2. Delete scope returns 0 when errors
Screenshot from 2022-09-05 21-22-20

ErykSol:
Nice catch on scope Error displaying twice- Fixed :)

PR #666, opened 2022-09-01
Unpipe input from topic when instance is terminated

patuwwy:
**What?**  
The problem with topics can be caused by extended instance lifetime. topic data remains piped to ended, but still existing instance.

- [x] All STH tests pass


alicja-gruzdz (approved):
Proper cleanup, thanks!

alicja-gruzdz (approved):

MichalCz (approved):

PR #667, opened 2022-09-01
Feature/telemetry label

Tatarinho:


packages/telemetry/src/adapters/loki.ts
```diff
@@ -26,6 +26,7 @@ export default class LokiAdapter implements ITelemetryAdapter {
                 new LokiTransport({
                     ...this.config,
                     replaceTimestamp: true,
+                    onConnectionError: (err) => console.error("Connection to Grafana not established", err)
```
patuwwy, line 29: actually it is connection to Loki.
please change to logger and just ```this.logger.error("Connection error")```

patuwwy (changes requested):

alicja-gruzdz (approved):
I'm waiting for the working endpoint to verify the stored information. Please do not merge until the verification is not completed, thanks.

I tested the error handling part - when the endpoint is unavailable the STH logs the info, so this part is working well.
Screenshot from 2022-09-05 09-39-36

packages/telemetry/src/adapters/loki.ts
```diff
@@ -26,6 +26,7 @@ export default class LokiAdapter implements ITelemetryAdapter {
                 new LokiTransport({
                     ...this.config,
                     replaceTimestamp: true,
+                    onConnectionError: (err) => this.logger.error("Connection error")
```
MichalCz, line 29:
```suggestion
                    onConnectionError: (err) => this.logger.error("Telemetry connection error")
```

packages/host/src/lib/host.ts
```diff
@@ -1054,6 +1055,12 @@ export class Host implements IComponent {
             this.telemetryAdapter = await getTelemetryAdapter(this.config.telemetry.adapter, this.config.telemetry);
             this.telemetryAdapter.logger.pipe(this.logger);
 
+            let ipAddress = require("ext-ip")();
+            ipAddress.on("ip", (ip: any) => {
+                this.ipvAddress = ip;
+            });
+            await ipAddress();
```
MichalCz, line 1062: @alicja-gruzdz @daro1337 @stokowski - I'm not sure about that, wouldn't we know what was the reporting IP anyway? And this module uses 3rd party services to get the IP...

alicja-gruzdz: Perhaps Loki has something, @Tatarinho is looking into it. 
Previously, @daro1337 suggested https://www.npmjs.com/package/public-ip but there were issues with it. It also uses external services and has deps.

alicja-gruzdz (commented):
Screenshot from 2022-09-06 21-18-16
Screenshot from 2022-09-06 21-19-01
Screenshot from 2022-09-06 21-20-37

packages/host/src/lib/host.ts
```diff
@@ -230,7 +231,7 @@ export class Host implements IComponent {
             this.logger.error("Setting telemetry failed");
         });
 
-        this.telemetryAdapter?.push("info", { message: "Host started", labels: { hostSize: this.hostSize } });
+        this.telemetryAdapter?.push("info", { message: "Host started", labels: { hostSize: this.hostSize, ip: this.ipvAddress} });
```
alicja-gruzdz, line 234: Could we please add also a sequence adapter? It should be easy since the sequence adapter information is read a bit lower at:
`const adapter = await initializeSequenceAdapter(this.config);`
Then we can move the Telemetry push for "Host started" just after it.

packages/host/src/lib/host.ts
alicja-gruzdz: There was an import issue with the library https://www.npmjs.com/package/public-ip [ERR_REQUIRE_ESM]. I'd merge this PR with the `ext-ip`, and then replace it if required.

alicja-gruzdz:
Merging to [feature/telemetry-label](https://github.com/scramjetorg/transform-hub/tree/feature/telemetry-label) and continued testing and additions (if required) will be on that branch.

PR #684, opened 2022-09-14
[SHN-854] add nodejs 18.x LTS

daro1337:
**What?**  
adds NodeJS 18.x (Upcoming LTS)

**Why?**  
@patuwwy requested to add  NodeJS v18.x

**Review checks:**

- [x] All STH tests pass


a-tylenda (approved):

patuwwy:
i have added one commit here, with change "localhost" -> "127.0.0.1" in tests but we should test it with https://github.com/scramjetorg/transform-hub/pull/687 instead

MichalCz:
Hmm... Node.js 14 EOL is in April 2023...

I'm ok with adding 18, but 14 is still something we should support.

We should also consider adding an option to choose specific version of Python and Node.

ErykSol (approved):

alicja-gruzdz (approved):

daro1337:
> Hmm... Node.js 14 EOL is in April 2023...
> 
> I'm ok with adding 18, but 14 is still something we should support.
> 
> We should also consider adding an option to choose specific version of Python and Node in runners.

@MichalCz  well, it's just security updates for v14. I can add it to CI matrix, but this may lengthen the duration of the CI. It's up to you

PR #689, opened 2022-09-15
(feat) transform-hub helm chart

daro1337:
**What?**  
Helm chart for transform-hub - easy deploy STH on Kubernetes

**Why?**  

Discord users request

**Usage:**
1. pull transform-hub repo
2. [Install Helm](https://helm.sh/docs/intro/install/)
3. Use your k8s cluster or setup test cluster using [minikube](https://minikube.sigs.k8s.io/docs/start/)
    - `minikube start --extra-config=apiserver.service-node-port-range=[8000-65535](tel:8000-65535)`
4. `helm upgrade --install sth charts/ -n default --force`
    - ` minikube service sth --url` # check service IP:port
In my case:
```bash
$ minikube service sth --url
http://192.168.49.2:8000
```
Then:
```bash
$ curl -s http://192.168.49.2:8000/api/v1/version |jq
{
  "service": "@scramjet/host",
  "apiVersion": "v1",
  "version": "0.28.3",
  "build": "8ad7715"
}

```

**How it works:**
Deploy STH to kubernetes cluster

- [ ] All STH tests pass


MichalCz:
@daro1337 could you move this to docs? I'm a bit worried that users could think that STH needs k8s to function and when they just want to run it as a process they could run away...

Otherwise it looks great! :)

alicja-gruzdz (approved):
Thanks for your help with minikube.
So far so good, tomorrow tbc.
Screenshot from 2022-09-15 17-40-14

Screenshot from 2022-09-15 17-44-56

daro1337:
@MichalCz Yes, I can, however, from my observations the standard in projects is the `charts/` directory in root directory.
IMO, docs define how to use our product

MichalCz:
This is my reasoning actually: users would think that this is something that needs k8s to run, and there are much simpler way to run it. If we put it in docs, we say: "this is a guide, adapt it to your situation", not "this is the intended way to run Scramjet Platform"

daro1337:
@MichalCz Okay, moved to `docs/guides/kubernetes/charts`

daro1337:
@alicja-gruzdz  I updated the PR description with a detailed instruction

docs/guides/kubernetes/charts/README.md
```diff
@@ -0,0 +1,13 @@
+## Exaple of deployment:
```
a-tylenda, line 1:
```suggestion
## Example of deployment:
```

MichalCz (approved):
Cool and great!

PR #687, opened 2022-09-14
Start host default on ::

patuwwy:
**What?**  
change "0.0.0.0" to "::"

**Why?**  
https://github.com/nodejs/node/issues/40537

**Usage:**

-
-
-

**How it works:**

-
-
-

**Review checks:**

These aspects need to be checked by the reviewer:

- [ ] Verify and confirm operation (please post a screenshot) 
- [ ] All STH tests pass
- [ ] All [Scramjet Cloud Platform](https://docs.scramjet.org/platform) tests pass
- [ ] Documentation is updated or no changes

PR #690, opened 2022-09-15
cli/tests, default to ipv4

patuwwy:
**What?**  
default to ipv4

**Review checks:**

- [x] All STH tests pass

### DIFF
--- a/packages/cli/src/bin/index.ts
+++ b/packages/cli/src/bin/index.ts
@@ -13,6 +13,7 @@ import { initConfig, profileConfig, profileManager } from "../lib/config";
 import { initPaths } from "../lib/paths";
 import chalk from "chalk";
 import { isProductionEnv } from "../types";
+import * as dns from "dns";
 
 const version = findPackage(__dirname).next().value?.version || "unknown";
 const CommandClass = completionMixin(commander).Command;
@@ -27,11 +28,15 @@ const initPlatform = async () => {
      * are provided in the profile configuration.
      * Do not set the default platform values when displaying the help commands.
      */
-    if (token && isProductionEnv(env) && middlewareApiUrl &&
+    if (
+        token &&
+        isProductionEnv(env) &&
+        middlewareApiUrl &&
         !process.argv.includes((program as any)._helpShortFlag) &&
-        !process.argv.includes((program as any)._helpLongFlag)) {
+        !process.argv.includes((program as any)._helpLongFlag)
+    ) {
         ClientUtils.setDefaultHeaders({
-            Authorization: `Bearer ${token}`
+            Authorization: `Bearer ${token}`,
         });
 
         await setPlatformDefaults();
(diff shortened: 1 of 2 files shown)


packages/cli/src/bin/index.ts
```diff
@@ -42,26 +47,33 @@ const initPlatform = async () => {
  * Start commander using defined config {@link Apple.seeds}
  */
 (async () => {
+    // https://nodejs.org/api/dns.html#dnssetdefaultresultorderorder
+    const { setDefaultResultOrder } = dns as unknown as { setDefaultResultOrder?: (param: string) => void };
+
+    if (setDefaultResultOrder) {
+        setDefaultResultOrder("ipv4first");
+    }
+
```
patuwwy, line 56: probably it needs to be in config

daro1337: yep :)

daro1337:
@patuwwy LGTM, looks like all tests passed especially on the node v18. Good job

daro1337 (approved):

ErykSol (approved):

alicja-gruzdz (commented):
just a question: if I understand correctly the behavior concerns `localhost`, right? like in https://github.com/nodejs/node/issues/40537 (this is for 17, I know but I wonder if it's this issue). I saw @daro1337 changed to 127.0.0.1 in https://github.com/scramjetorg/transform-hub/pull/684 and you changed it in the STH config in https://github.com/scramjetorg/transform-hub/pull/687

Is setting here setDefaultResultOrder to IPv4 a safeguard in case `localhost` would be used in BDD/CLI? because if so, would we need it in STH (Runner connection)?

daro1337:
@alicja-gruzdz  https://github.com/nodejs/node/issues/40537#issuecomment-948047193
In some tests we're using `localhost`
https://github.com/scramjetorg/transform-hub/blob/878903380c3e4e0d0fd055dab9e3ed48ba1c5510/bdd/step-definitions/e2e/host-steps.ts#L119
https://github.com/scramjetorg/transform-hub/blob/878903380c3e4e0d0fd055dab9e3ed48ba1c5510/bdd/step-definitions/e2e/host-steps.ts#L175

```bash
$ host localhost
localhost has address 127.0.0.1
localhost has IPv6 address ::1
```

So if we start STH only on IPv4 interface, and our `si` based on node v17+ resolve `localhost` to `::1` we're going to get Connection refused.

alicja-gruzdz:
> 

Hey @daro1337 :) thanks for clarifying. I was wondering what if a user is on Node 17+ but specifies as Host address localhost. Whether the Runner will connect to STH in such case. However, probably users who change such settings are developers and will be aware of this issue.

daro1337:
@alicja-gruzdz IMO it depends on `--runtime-adapter`, in k8s we have `--k8s-sth-pod-host ` in  docker AFAIK we're creating `transformhub0` bridge interface. I don't have knowledge about process adapter

patuwwy:
> @alicja-gruzdz [nodejs/node#40537 (comment)](https://github.com/nodejs/node/issues/40537#issuecomment-948047193) In some tests we're using `localhost`
> 
> https://github.com/scramjetorg/transform-hub/blob/878903380c3e4e0d0fd055dab9e3ed48ba1c5510/bdd/step-definitions/e2e/host-steps.ts#L119
> 
> 
> https://github.com/scramjetorg/transform-hub/blob/878903380c3e4e0d0fd055dab9e3ed48ba1c5510/bdd/step-definitions/e2e/host-steps.ts#L175
> 

this has been changed in #696

alicja-gruzdz (approved):

PR #697, opened 2022-09-19
[CLI] fix loading of broken profile files

ErykSol:
What:
This PR fixes loading of broken config files in CLI, and adds little clean up to config files.

Why:
If user manualy changed profile files, or profile was damaged during test crash, CLI had edge cases when it wasn't able to start properly. This PR exclude this possibility, and adds check on using broken profile file.

How it works::
create manualy empty profile in yourHomeDir/.si/profiles/testProfile.json
`si c pr use testProfile`
Earlier this would crash CLI for further use, now it reset profile to run properly on default config.

- [x] All STH tests pass


packages/cli/src/lib/errorHandler.ts
```diff
@@ -12,9 +13,25 @@ async function jsonize(err: ClientError) {
...
+        displayError(error, debug);
+    }
+    return { format, debug };
+};
+
 export const errorHandler = async (err: Error) => {
     process.exitCode = getExitCode(err);
-    const { log: { format, debug } } = profileConfig.getConfig();
+
+    const { format, debug } = getLogParams();
```
ErykSol, line 34: This funcion asserts that params are read properly even on broken config files.

alicja-gruzdz (approved):
++ for good refactor, the code is well structured now.

PR #700, opened 2022-09-21
Add args field to instance, change sequenceArgs to args in sequences

ErykSol:
This PR:
1) changes "sequenceArgs" to field "args" in instance details. New structure looks like this (i.e. GET instance/id ):
```
{ ...
  args?: any[], <--args provided by user on start of instance
  sequenceInfo: {
     args?: any[], <--args provided in package.json of sequence
  }
}
```

2) Adds optional field "args" in sequence details. New structure looks like this (i.e. GET sequence/id ):
{ ...
  config: {
    args?: any[]  <--args provided in package.json of sequence
  }
}

3) Changes returned required fields:
 -"author"
 -"keywords"
 -"repository"
 from package.json to optional fields. This fields are optional in package.json standard, and should be kept that way for consistency. This also allow us to get rid of empty values like: keywords: [], author: "" etc. returned in console causing redundant amount of text which doesn't bring any info.

4) Changes order of args provided to instance.
Previous: if args provided in req.body- pass them on instance start
Now:  if args provided in req.body- pass them on instance start, if not provided but exists in package.json of sequence use args from sequence.

**Review checks:**

- [x] Verify and confirm operation (please post a screenshot) 
- [x] All STH tests pass


packages/cli/src/lib/commands/sequence.ts
```diff
@@ -107,7 +107,7 @@ export const sequence: CommandDefinition = (program) => {
         .option("--limits <json-string>", "Instance limits")
         .description("Start the Sequence with or without given arguments")
         .action(async (id, { configFile, configString, outputTopic, inputTopic, args: argsStr, limits: limitsStr }) => {
-            const args = sequenceParseArgs(argsStr);
+            const args = argsStr ? sequenceParseArgs(argsStr) : undefined;
```
patuwwy, line 110:
```suggestion
            const args = argsStr && sequenceParseArgs(argsStr);
```

patuwwy (approved):

packages/cli/src/lib/commands/sequence.ts
ErykSol: Good catch- thanks

agne6ka (approved):
Woah, that was tricki 😄, thanks for the description how to do it 🙌 ! It was much easier to test it for such a 🐴 like me 😉 

### Seq with no args:
image
1. No args has been passed when inst starts:
image
2. Args has been passed when inst starts:
image
### Seq with args:
image
1. No args has been passed when inst starts (so it started with a defaults):
image
2. Args has been passed when inst starts:
image

PR #713, opened 2022-10-10
(helm) sth support configMap

daro1337:
**What?**  
helm chart for STH -  added `config` block with STH configurtion for [IaC mode ](https://github.com/scramjetorg/transform-hub/pull/239)

**Why?**  
allow to configure STH in IaC mode 

We can start STH from config file e.g  `--startup-config=/config/config.json`

**Usage:**
1. pull transform-hub repo
2. [Install Helm](https://helm.sh/docs/intro/install/)
3. Use your k8s cluster or setup test cluster using [minikube](https://minikube.sigs.k8s.io/docs/start/)
    - `minikube start --extra-config=apiserver.service-node-port-range=[8000-65535]`
4. Edit values.yaml :
```bash
## ConfigMaps
configDir: /config/

config:
    config.json: |
      {
        "sequences": [
          { "id": "default", "args": [],  "appConfig": {} }
        ]
      }
```
5. `helm upgrade --install sth charts/ -n default --force`

Then check if config exists:
```bash
$ kubectl exec -ti sth-0 sth -- cat /config/config.json

{
  "sequences": [
    { "id": "default", "args": [],  "appConfig": {} }
  ]
}
```

**How it works:**
Allows to add helper containers

- [ ] All STH tests pass

### DIFF
--- a/docs/guides/kubernetes/charts/templates/sth-config.yaml
+++ b/docs/guides/kubernetes/charts/templates/sth-config.yaml
@@ -1,0 +1,17 @@
+{{- if .Values.config }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: sth-cm
+  labels:
+    app.kubernetes.io/name: {{ include "sth.name" . }}
+    helm.sh/chart: {{ include "sth.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+data:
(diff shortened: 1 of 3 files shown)


alicja-gruzdz (approved):
It will be a super useful option for starting STH with existing Sequences on kube! :raised_hands: 
(I remember we still need to debug one thing on the STH backed but the chart is excellent).

thanks for your help with resolving the minikube issue & all!

Screenshot from 2022-10-11 15-18-45

PR #784, opened 2023-01-13
Align Sequences root dir name for all Adapters

alicja-gruzdz:
**What?**  

The directory where Sequences are saved when running STH with Kubernetes Adapter can now be specified with the option:
`--sequences-root <path>`
The same option is currently used for Process Adapter.

The option `--k8s-sequences-root`  will be deprecated in the near future in SCP.

**Why?**  
Unifying usage in the context of persistence Sequences feature..

**Usage:**

-
-
-

**How it works:**

-
-
-

**Review checks:**

- [ ] All STH tests pass


stokowski (approved):

daro1337:
@alicja-gruzdz  can you update: https://github.com/scramjetorg/transform-hub/blob/devel/docs/guides/kubernetes/charts/templates/sth-statefulset.yaml#L99

daro1337 (changes requested):
https://github.com/scramjetorg/transform-hub/blob/devel/docs/guides/kubernetes/charts/templates/sth-statefulset.yaml#L99

daro1337:
@alicja-gruzdz  actually, It might be out of scope, but is it possible to introduce ` -D / --sequences-root <path>"` to kubernetes adpater with fallback to `homedir(), ".scramjet_sequences`?

It's argument for process adapter:
```bash
    .option("-D, --sequences-root <path>", "Only works with --runtime-adapter='process' option. Where should ProcessSequenceAdapter save new sequences")
```

daro1337:
https://github.com/scramjetorg/transform-hub/pull/785

alicja-gruzdz:
> homedir(), ".scramjet_sequences

Hey Darek,

sure. 

When I had a closer look at these options I found that there is an option to override the sequence directory for Kubernetes Adapter already:
https://github.com/scramjetorg/transform-hub/blob/devel/packages/sth/src/bin/hub.ts#L46
`sth  --runtime-adapter='kubernetes' --k8s-sequences-root="some/seq/dir/for/k8s/adapter"`

and from what I'm checking it works as expected. If no value is provided it defaults to (now) `homedir(), ".scramjet_sequences `.

However, I do not remember (and neither does @patuwwy) why we have two different options to configure the sequence directory: `--sequences-root` vs `--k8s-sequences-root ` - do you perhaps recall?
I could unify them, but I would like first to make sure it won't break anything.

alicja-gruzdz:
@daro1337 I introduced the changes we discussed. The order of selecting the sequence directory for k8s adapter now is:
1. --sequences-root
2. --k8s-sequences-root
3. homedir() + .scramjet_sequences

I dry tested it by printing the config in Kubernetes Adapter while running STH. Please let me know if you spot any issues, thanks.

daro1337:
Can you update description for `-D` argument? Also `--k8s-sequences-root` should be considered as detracted in the near feature, and we should point out to it's description IMO :)
```bash
option("-D, --sequences-root <path>", "Only works with --runtime-adapter='process' option. Where should ProcessSequenceAdapter save new sequences")
```

daro1337 (approved):

a-tylenda (approved):

PR #792, opened 2023-01-25
Add crash log feature to Kube Adapter

MichalCz:
**What?**  

This adds a crash log to Kubernetes Adapter. When a sequence fails the crash log 
(i.e. last 100 lines of std out/err) will be extracted./

**Why?**  

Currently this is not implemented and makes debugging difficult.

**Usage:**

- Send a crashing sequence using `si seq deploy`
- When the sequence crashed get `si inst logs -`
- The Crashlog should have content

**Review checks:**

- [ ] All STH tests pass


patuwwy (approved):

alicja-gruzdz:
I'm adding the do not merge label, because I'm debugging this feature with an example failing Sequence https://github.com/scramjetorg/reference-apps/blob/main/js/exception-test/src/index.ts (thanks Patryk) and it causes handing request when starting the Sequence.

MichalCz (commented):
Ok.

MichalCz (commented):
Ok

PR #795, opened 2023-01-26
Autodetect runtime adapter when not specifically set

MichalCz:
**What?**  

Automatically choose process adapter when runtime adapter is set to `detect` or not set at all.

This will try to use docker and fall-back if docker is not found.

**Why?**  

Most users trying this out for the first time might not want to install docker. This will suit everyone.

**Usage:**

- start on a machine without docker - process adapter should be chosen, no error should appear.
- start on a machine with docker - docker adapter should be chosen, no error should appear.
- choosing any adpater should work as previously

**How it works:**

- configuration is set to `detect` by default
- the docker adapter checks if `dockerode.info()` command doesn't fail
- if so, it chooses docker

**Review checks:**

- [ ] All STH tests pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@scramjet/transform-hub",
-  "version": "0.31.4",
+  "version": "0.32.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@scramjet/transform-hub",
-      "version": "0.31.4",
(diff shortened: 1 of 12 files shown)


patuwwy (approved):

patuwwy:
CI issues caused by npm@9. Please rebase to fix.

alicja-gruzdz:
excellent, much-needed feature.

alicja-gruzdz (approved):

PR #807, opened 2023-02-21
cli init sequence

ErykSol:
**What?**  
Add command `si init sequence` to CLI

**Why?**  
To make life of a user simpler...

**Usage:**
In sequence folder run 
`si init sequence js`
`si init sequence py` (same as `si init sequence`)
`si init sequence ts`

run `si init sequence -h` to see possible arguments

**How it works:**
It spawns `npm init scramjetorg/sequence` with proper param as childprocess

- [x] All STH tests pass
- [x] All [Scramjet Cloud Platform](https://docs.scramjet.org/platform) tests pass

Waiting for:
https://github.com/scramjetorg/create-sequence/pull/8

### DIFF
--- a/packages/cli/src/lib/commands/init.ts
+++ b/packages/cli/src/lib/commands/init.ts
@@ -1,3 +1,5 @@
+import { Argument } from "commander";
+import { spawnSync } from "child_process";
 import { CommandDefinition, ExtendedHelpConfiguration } from "../../types";
 
 export const init: CommandDefinition = (program) => {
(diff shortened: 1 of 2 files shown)


alicja-gruzdz:
thanks, it will simplify the Sequence creation for the new users (and not only). I'll review tomorrow with @RayNawfal

a-tylenda:
A description is missing for the `si init` command:
image

a-tylenda:
Verified for all 3 languages, works as expected 👍🏼

packages/cli/src/lib/commands/init.ts
```diff
@@ -1,3 +1,5 @@
+import { Argument } from "commander";
+import { spawnSync } from "child_process";
```
MichalCz, line 2: Eryk, why use spawnSync? Async spawn would be better.

MichalCz (approved):
This is ok, but perhaps take a look at the comment.

ErykSol:
> This is ok, but perhaps take a look at the comment.

It's due to the nature of commander synchronous work- once entered the command handling no other tasks are running under the hood so there was no point to make this async

alicja-gruzdz (approved):

PR #843, opened 2023-04-13
Fix for Hub Client error on Broken Streams

MichalCz:
**What?**  

The PR solves the issue of STH crashing when a sequence using `this.hub` is killed. The solution was to properly handle errors on disconnection of the additional stream in the connection between the runner and the hub.

**Why?**  

Gotta fix bugs.

**Review checks:**

- [x] All STH tests pass

### DIFF
--- a/packages/host/src/lib/csi-controller.ts
+++ b/packages/host/src/lib/csi-controller.ts
@@ -556,6 +556,9 @@ export class CSIController extends TypedEmitter<Events> {
             this.createInstanceAPIRouter();
 
             this.bpmux = new BPMux(streams[8]);
+            this.bpmux.on("error", (e: any) => {
+                throw new Error("BPMux Carrier Error:" + e.message);
+            });
             this.bpmux.on("peer_multiplex", (socket: Duplex, _data: any) => this.hostProxy.onInstanceRequest(socket));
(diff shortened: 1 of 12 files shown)


packages/host/src/lib/csi-controller.ts
```diff
@@ -556,6 +556,10 @@ export class CSIController extends TypedEmitter<Events> {
             this.createInstanceAPIRouter();
 
             this.bpmux = new BPMux(streams[8]);
+            this.bpmux.on("error", (e: any) => {
+                this.logger.warn("Instance client multiplex connection errored", e.message);
+                streams[8]?.end();
```
alicja-gruzdz, line 561: the error was thrown then hub client send request to kill Instance, right? why do we need to end the hub client stream in such case (and only this one)?

alicja-gruzdz (approved):

packages/host/src/lib/csi-controller.ts
MichalCz: This also ends all the requests sent over the connections. This will clean up the connections and get the allocated memory back.

alicja-gruzdz: clear, thanks

PR #861, opened 2023-05-17
New Cli command that allows restarting instance

piotrek6641:
**What?**  
```
si seq start XX
{ id: SEQ_ID }
```
```
si inst restart SEQ_ID
```

restarted instance keeps id, input topic, output topic if set.
 

**Why?**  
`si seq start -` + `si inst kill -` + `si seq start - params` = `si inst restart -`

**Usage:**

-
-
-

**How it works:**

-
-
-

**Review checks:**

- [ ] All STH tests pass

### DIFF
--- a/packages/cli/src/lib/helpers/instance.ts
+++ b/packages/cli/src/lib/helpers/instance.ts
@@ -3,7 +3,6 @@ import { getHostClient, getInstance } from "../common";
 import { getInstanceId, sessionConfig } from "../config";
 import { SequenceClient } from "@scramjet/api-client";
 import { defer } from "@scramjet/utility";
-//import { ConfigFileDefault } from "@scramjet/utility";
 
 /**
  * Wraper for kill instance method with config safety checks.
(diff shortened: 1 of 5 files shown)


piotrek6641:
Need to fix this before: https://github.com/scramjetorg/transform-hub/issues/862

piotrek6641:
Will also need to have informations about topics in instanceInfo so they will be kept when restarting

packages/cli/src/lib/helpers/instance.ts
```diff
@@ -23,3 +25,38 @@ export const instanceKill = async (
...
+            await getInstance(instanceId).getInfo();
+            await defer(100);
+        } catch (e:any) {
+            if (e.status === "404") {
+                ready = true;
+                break;
+            }
+        }
+    }
+    if (typeof sequenceId === "string") {
```
patuwwy, line 51:
```suggestion

    if (typeof sequenceId === "string") {
```

piotrek6641:
#644

patuwwy (approved):

alicja-gruzdz (approved):

patuwwy (approved):

PR #877, opened 2023-06-02
Test for issue with API Server path handling

MichalCz:
**What?**  

API Server accepts paths strictly as their ASCII representation and that was the intended way, but using `+` in the path breaks the route.

What should be happenning is that either:

* querystring unescape should be applied
* `+` char should be treated as a non-special case

Included 

**Why?**  

However this has no current impact, path handling should work consistently.

Running this test will show the issue:

```bash
$ yarn test:packages -s @scramjet/api-server
```

**Review checks:**

- [ ] All STH tests pass


alicja-gruzdz (approved):

alicja-gruzdz:
Will be merged with the test passes.

PR #881, opened 2023-06-06
fix/proper-permissions-after-unpack

tomekcrm:
**What?**  

Proper permissions were assigned to the unpacked files to enable their processing by the sequence.

**Why?**  

Improper permission assignment prevented the reading of the .wav file, resulting in the interruption of the sequence.

**Usage:**

- Run STH from the "fix/proper-permissions-after-unpack" branch 
- load and run the py-transcript sequence to verify the correctness of STH's operation.

**Review checks:**

- [*] Verify and confirm operation (please post a screenshot) 
- [ ] All STH tests pass

[logs from testing](https://github.com/scramjetorg/transform-hub/files/11662541/py-transcript-logs.log)

### DIFF
--- a/packages/python-runner/unpack.sh
+++ b/packages/python-runner/unpack.sh
@@ -2,4 +2,4 @@
 
 set -e
 
-tar zxf - -C /package && touch /package/.ready || touch $PACKAGE_DIR/.fail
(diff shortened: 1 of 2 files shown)


alicja-gruzdz (approved):
sweet, thanks!

Issue #850, opened 2023-04-24
CLI: Unable to set platform defaults


alicja-gruzdz:
**Describe the bug**
Setting platform defaults is triggered even when the config is printed or spaces/hubs display and returns an unclear error message.

Example 1:
![Screenshot from 2023-04-24 11-44-42](https://user-images.githubusercontent.com/8177865/233964848-6a7920fc-6788-488c-af85-942fe34c923a.png)

Example 2:
![Screenshot from 2023-04-24 12-04-37](https://user-images.githubusercontent.com/8177865/233965988-20bdfe00-86b4-4c48-b0d2-b1b435b2a80a.png)

and there are more such occurrences when executing other commands.

**Steps to Reproduce**
1. Set the active CLI profile to "production" env when the spaceId/hubId is not correct.
2. Execute config print.

**Expected behavior**
The defaults must be set when changing Space or Hub (si space use or si hub use). When the Space or Hub is not reachable the error message must be clear and there must be instructions on how to fix it (e.g. checking space and hubs with ls and setting the active one with `si space use` or `si hub use`)

alicja-gruzdz:
Space ID is still not updated after the CLI config is set:
![image](https://github.com/scramjetorg/transform-hub/assets/8177865/68fd69e5-95cb-41e2-8c24-54ce51720528)

PR #883, opened 2023-06-15
Fix: Unable to set platform defaults

piotrek6641:
**What?**  
Compare whether LastSpaceID matches SpaceID from MiddlewareManager
**Why?**  
When changing tokens the CLI session was holding old Manager ID which was leading to receiving error 403
https://github.com/scramjetorg/transform-hub/issues/850

**Usage:**

- Have 2 accounts on a platform 
- Change account using generated token on a website
-  Try listing instances or make any other operation.

**Review checks:**

- [ ] All STH tests pass

### DIFF
--- a/packages/cli/src/lib/platform/common.ts
+++ b/packages/cli/src/lib/platform/common.ts
@@ -36,13 +36,16 @@ export const getMiddlewareClient = (): MiddlewareClient => {
 };
 
 export const setPlatformDefaults = async () => {
+    const middlewareClient = getMiddlewareClient();
+    const middlewareManager = await middlewareClient.getManagers();
+
     const { lastSpaceId, lastHubId } = sessionConfig.get();
 
     if (lastSpaceId || lastHubId) {
-        return false;
+        if (middlewareManager[0]?.id === lastSpaceId) {
+            return false;
+        }
     }
-
-    const middlewareClient = getMiddlewareClient();
 
     try {
         const managers = await middlewareClient.getManagers();
--- a/packages/cli/src/lib/platform/common.ts
+++ b/packages/cli/src/lib/platform/common.ts
@@ -37,19 +37,17 @@ export const getMiddlewareClient = (): MiddlewareClient => {
 
 export const setPlatformDefaults = async () => {
     const middlewareClient = getMiddlewareClient();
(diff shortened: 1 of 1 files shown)


packages/cli/src/lib/platform/common.ts
```diff
@@ -36,14 +36,17 @@ export const getMiddlewareClient = (): MiddlewareClient => {
...
 
 export const setPlatformDefaults = async () => {
+    const middlewareClient = getMiddlewareClient();
+    const middlewareManager = await middlewareClient.getManagers();
+
     const { lastSpaceId, lastHubId } = sessionConfig.get();
 
     if (lastSpaceId || lastHubId) {
-        return false;
+        if (middlewareManager[0].id === lastSpaceId) {
```
patuwwy, line 45:
```suggestion
        if (middlewareManager[0]?.id === lastSpaceId) {
```

patuwwy (approved):

alicja-gruzdz (approved):
Verified by switching between two diffrent accounts using the same default profile and executing a couple of listing commands - it worked fine, thanks. Let's merge it and I will release a new STH this week.

packages/cli/src/lib/platform/common.ts
```diff
@@ -36,14 +36,17 @@ export const getMiddlewareClient = (): MiddlewareClient => {
 };
 
 export const setPlatformDefaults = async () => {
+    const middlewareClient = getMiddlewareClient();
+    const middlewareManager = await middlewareClient.getManagers();
```
patuwwy, line 40: managers

patuwwy: and can be reused at line 51 without requesting MW second time

patuwwy (changes requested):
see comments

packages/cli/src/lib/platform/common.ts
piotrek6641: fixed with [91f7afc](https://github.com/scramjetorg/transform-hub/pull/883/commits/91f7afc83a4bfcf7ba2aaf7c67c5c047af3abf9d)

patuwwy (approved):

Issue #889, opened 2023-06-23
[SDK] Add method `getSequenceId(seq-name)` to HostClient class


a-tylenda:
## New method implementation

Please add a new method called `getSequenceId(seq-name)` to the `HostClient` class. The method should take the sequence name from `sequenceInfo.config.name` as an argument.

Sequence info:
{
  id: '7a7e38f4-2f3b-4e49-b05a-cd3a8877c0fa',
  config: {
    type: 'process',
    engines: {},
    entrypointPath: 'index.js',
    version: '0.22.0',
    name: 'new-sequence',
    id: '7a7e38f4-2f3b-4e49-b05a-cd3a8877c0fa',
    sequenceDir: '/home/user/.scramjet_sequences/7a7e38f4-2f3b-4e49-b05a-cd3a8877c0fa',
    description: '',
    author: 'Scramjet <open-source@scramjet.org>',
    repository: {
      type: 'git',
      url: 'https://github.com/scramjetorg/transform-hub.git'
    },
    language: 'js',
    packageSize: 18952
  },
  instances: []
}

This method should return Sequence ID as a string. 

However, to ensure the effectiveness and coherence of this method, it is necessary to implement a modification that prevents the user from sending to the Hub a Sequence that has an identical name in `package.json`.

PR #919, opened 2023-07-25
Feat/get sequence

piotrek6641:
What: 
Finding list of ID's for a sequences with a given name.
Clickup:
https://app.clickup.com/t/24308805/VDM-1298

**How to test:**
- si seq deploy packages/forever.tar.gz
- ts-node packages/cli/src/bin/index.ts sequence list -n @scramjet/js-forever

**Review checks:**

- [ ] All STH tests pass

### DIFF
--- a/packages/api-client/src/host-client.ts
+++ b/packages/api-client/src/host-client.ts
@@ -32,6 +32,19 @@ export class HostClient implements ClientProvider {
         return this.client.get<STHRestAPI.GetSequencesResponse>("sequences");
     }
 
+    async getSequenceId(sequenceName: string) {
+        const sequenceList = await this.client.get<STHRestAPI.GetSequencesResponse>("sequences");
+
+        const result = sequenceList.filter((sequence) => {
+            return sequence.config.name === sequenceName;
+        });
+
+        if (result.length !== 1) {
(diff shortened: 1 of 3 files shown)


piotrek6641:
Needs validation of name when new sequence is being uploaded/ and functionality moved to a different file

piotrek6641:
This PR doesn't include method to prevent user from sending multiple sequences with the same name as stated in https://github.com/scramjetorg/transform-hub/issues/889 . After discussion with @alicja-gruzdz we currently don't plan to implement this functionality

ErykSol (changes requested):

ErykSol (approved):

alicja-gruzdz:
Then successful it looks great:
Screenshot from 2023-07-28 17-01-51

alicja-gruzdz:
Then it does not find the Sequence it throws an error which we need to handle:
Screenshot from 2023-07-28 17-03-55

alicja-gruzdz:
> Then it does not find the Sequence it throws an error which we need to handle: Screenshot from 2023-07-28 17-03-55

@piotrek6641 checked it is due to running it with ts-node

alicja-gruzdz (approved):

PR #925, opened 2023-08-04
fix for 403 error

piotrek6641:
Temporary fix to prevent 403 error when having expired token/ account removed

**Clickup Task:** 
https://app.clickup.com/t/24308805/VDM-1017

**How it works:**

-
-
-

**Review checks:**

These aspects need to be checked by the reviewer:

- [ ] Verify and confirm operation (please post a screenshot) 
- [ ] All STH tests pass
- [ ] All [Scramjet Cloud Platform](https://docs.scramjet.org/platform) tests pass
- [ ] Documentation is updated or no changes

PR #930, opened 2023-08-11
Fix CLI not to read space state when it's not needed

MichalCz:
**What?**  

Fixes CLI not to check space connectivity on every single setting.

**Why?**  

If you're changing your config, there's no point in checking if the space is connected

**Usage:**

-
-
-

**Clickup Task:** 

**How it works:**

-
-
-

**Review checks:**

- [ ] All STH tests pass

### DIFF
--- a/packages/cli/src/lib/commands/instance.ts
+++ b/packages/cli/src/lib/commands/instance.ts
@@ -6,6 +6,7 @@ import { getInstanceId, profileManager, sessionConfig } from "../config";
 import { displayEntity, displayObject, displayStream } from "../output";
 import { ClientError } from "@scramjet/client-utils";
 import { Option } from "commander";
+import { initPlatform } from "../platform";
 /**
  * Initializes `instance` command.
  *
@@ -14,6 +15,7 @@ import { Option } from "commander";
 export const instance: CommandDefinition = (program) => {
     const instanceCmd = program
         .command("instance [command]")
+        .hook("preAction", initPlatform)
         .addHelpCommand(false)
         .configureHelp({ showGlobalOptions: true })
         .alias("inst")
--- a/packages/cli/src/lib/commands/sequence.ts
+++ b/packages/cli/src/lib/commands/sequence.ts
@@ -12,6 +12,7 @@ import { isDevelopment } from "../../utils/envs";
 import { resolve } from "path";
 import { sequenceDelete, sequencePack, sequenceParseArgs, sequenceSendPackage, sequenceStart } from "../helpers/sequence";
 import { ClientError } from "@scramjet/client-utils";
+import { initPlatform } from "../platform";
 
 /**
  * Initializes `sequence` command.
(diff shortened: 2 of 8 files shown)


packages/cli/src/lib/platform/common.ts
```diff
@@ -71,3 +74,31 @@ export const setPlatformDefaults = async () => {
...
+    env: configEnv,
+    middlewareApiUrl: string
+) =>
+    token &&
+    isProductionEnv(env) &&
+    middlewareApiUrl &&
+    !process.argv.includes(program._helpShortFlag) &&
+    !process.argv.includes(program._helpLongFlag);
+
+export const initPlatform = async (program: Command) => {
```
ErykSol, line 91: I would add extra check, just in case:
`if (!isProductionEnv(profileConfig.env)) return;`

packages/cli/src/lib/commands/instance.ts
```diff
@@ -14,6 +15,7 @@ import { Option } from "commander";
 export const instance: CommandDefinition = (program) => {
     const instanceCmd = program
         .command("instance [command]")
+        .hook("preAction", initPlatform)
```
ErykSol, line 18: This doesn't require initPlatform while on development env

packages/cli/src/lib/commands/sequence.ts
```diff
@@ -21,6 +22,7 @@ import { ClientError } from "@scramjet/client-utils";
 export const sequence: CommandDefinition = (program) => {
     const sequenceCmd = program
         .command("sequence")
+        .hook("preAction", initPlatform)
```
ErykSol, line 25: Same here: doesn't require initPlatform while on development env

packages/cli/src/lib/commands/topic.ts
```diff
@@ -27,6 +28,7 @@ export const topic: CommandDefinition = (program) => {
 
     const topicCmd = program
         .command("topic")
+        .hook("preAction", initPlatform)
```
ErykSol, line 31: And here: doesn't require initPlatform while on development env

ErykSol (changes requested):
Small fixes and all is good.

ErykSol:
Please check this PR, its quick fix for previous problem, but still nice to have in your solution:
https://github.com/scramjetorg/transform-hub/pull/925/files

packages/cli/src/lib/platform/common.ts
MichalCz: I think this depends on which API we're calling - if we want to use topics on local, yes, but if we want to list hubs, that may be different.

I'll look into this.

piotrek6641: implemented @ErykSol solution, resolving all conversations

patuwwy (approved):

ErykSol (approved):

Issue #915, opened 2023-07-21
[TOPIC] Option `--content-type` in `si topic get <topicName>` does not work


a-tylenda:
**Describe the bug**
When I try to use other content type then default it id not working but only when initiating topic with `si topic get` command:

**Steps to Reproduce**
1. Create a topic by executing this command:
```bash
si topic get newTopic -t application/octet-stream
```

2. Then list topics
```bash
$ si topic ls
[
  {
    contentType: 'application/x-ndjson',
    localProvider: '',
    topic: 'newTopic',
    topicName: 'newTopic'
  }
]
```
The content type stays unchanged. Should be `application/octet-stream`.

**Expected behavior**
The content type should be possible to be overwritten when using `--content-type` or `-t` option

**Version:**

STH version: 0.35.0
SI version: 0.35.0
node version: v18.15.0
os: Ubuntu 20.4

ErykSol:
Checked on sth 35.2. Working

Issue #914, opened 2023-07-21
[TOPIC] Default content type inconsistency 


a-tylenda:
**Describe the bug**
The content type of a topic is not the same when initializing the topic in a 3 different ways.

**Steps to Reproduce**
I initiated a topic in a 3 different ways:

1. `si topic create createdTopic`

2. `si topic send sentTopic`

3. `si topic get getTopic`

After that, I listed all topics:

`si topic ls`

```json
[
  {
    contentType: 'text/plain',
    localProvider: '',
    topic: 'createdTopic',
    topicName: 'createdTopic'
  },
  {
    contentType: 'text/plain',
    localProvider: '',
    topic: 'sentTopic',
    topicName: 'sentTopic'
  },
  {
    contentType: 'application/x-ndjson',
    localProvider: '',
    topic: 'getTopic',
    topicName: 'getTopic'
  }
]
```
**Expected behavior**
Default `contentType: 'application/x-ndjson'` is set when initiating a topic by GET request. Should be text/plain as `si topic get --help` says:

```bash
$ si topic get -h
Current profile: local
Usage: si topic get [options] <topic-name>

Get data from topic

Options:
  -t, --content-type [content-type]  Specifies type of data in topic (choices: "text/x-ndjson", "application/x-ndjson", "text/plain", "application/octet-stream", default: "text/plain")
  -h, --help
```

**Version (please complete the following information):**
STH version: 0.35.0
SI version: 0.35.0
node version: v18.15.0
os: Ubuntu 20.4

**Additional information like logs, screenshots etc.**

ErykSol:
Tested on STH 35.2 Working

PR #932, opened 2023-08-21
fix/topic streaming bug

MichalCz:
**What?**  

- Fix CLI not to read space state when it's not needed
- Fix API server disconnection tracking

**Why?**  

Topics were wrongly redirected to new instances on every request. This fixes the issue.

Also makes topic writers queue up after each other - first topic writer needs to be consumed, before others can start writing.

Also audit stream logs were turned off - these cause a loop of messages.

**Usage:**

In separate terminals:

```bash
yarn start:dev
cat /dev/random | hexdump | head -n 10 | si topic send mytopic
cat /dev/random | hexdump | head -n 10 | si topic send mytopic # yes, again
# wait a little - above terminals should wait and topics should wait
si topic get mytopic
# topics should be consumed only after this command
```

**Clickup Task:** 

**How it works:**

- see consumePipe in sd-adapter.
- see changes in api-server (pause)

**Review checks:**

- [ ] All STH tests pass

### DIFF
--- a/packages/host/src/lib/auditor.ts
+++ b/packages/host/src/lib/auditor.ts
@@ -93,7 +93,8 @@ export class Auditor {
         const opCode = this.getOpCode(req);
         const requestorId = this.getRequestorId(req);
 
-        this.logger.trace("Requestor, tx, rx", requestorId, req.auditData.rx, req.auditData.tx);
+        // Logging this generates a loop of traffic
+        // this.logger.trace("Requestor, tx, rx", requestorId, req.auditData.rx, req.auditData.tx);
 
         if (opCode !== OpRecordCode.NOT_PROCESSABLE) {
             this.write({
--- a/packages/host/src/lib/serviceDiscovery/sd-adapter.ts
+++ b/packages/host/src/lib/serviceDiscovery/sd-adapter.ts
@@ -73,9 +73,13 @@ export class ServiceDiscovery {
      */
     createTopicIfNotExist(config: DataType) {
         const topicName = config.topic;
-        const topic = this.topicsController.get(topicName); // TODO: sprawdzanie content Type
+        const topic = this.topicsController.get(topicName);
 
         if (topic) {
+            if (topic.contentType !== config.contentType) {
+                this.logger.error("Content-type mismatch, existing and requested ", topic.contentType, config.contentType);
+                throw new Error("Content-type mismatch");
+            }
             this.logger.debug("Topic routed:", config);
             return topic;
         }
--- a/packages/host/src/lib/serviceDiscovery/sd-adapter.ts
+++ b/packages/host/src/lib/serviceDiscovery/sd-adapter.ts
@@ -53,13 +53,6 @@ export class ServiceDiscovery {
         return this.topicsController.get(id);
     }
 
-    createTopic(id: TopicId, contentType: ContentType) {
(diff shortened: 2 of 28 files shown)


packages/host/src/lib/serviceDiscovery/topic.ts
```diff
@@ -41,18 +52,40 @@ export class Topic extends Duplex implements TopicHandler {
...
     origin() { return this._origin; }
 
-    _read(_size: number): void { }
-    _write(chunk: any, encoding: BufferEncoding, callback: (error?: Error | null | undefined) => void): void {
+    acceptPipe(rdble: Readable) {
+        this._pipeQueue.push(rdble);
+        this.consumePipe();
+    }
+
+    consumePipe() {
```
ErykSol, line 60: This looks like it shouldn't be here. Duplicates logic from post topic where we call `.on("end")` anyway.

packages/host/src/lib/auditor.ts
```diff
@@ -93,7 +93,8 @@ export class Auditor {
         const opCode = this.getOpCode(req);
         const requestorId = this.getRequestorId(req);
 
-        this.logger.trace("Requestor, tx, rx", requestorId, req.auditData.rx, req.auditData.tx);
+        // Logging this generates a loop of traffic
```
ErykSol, line 96: YAGNI

packages/host/src/lib/serviceDiscovery/sd-adapter.ts
```diff
@@ -53,13 +53,6 @@ export class ServiceDiscovery {
         return this.topicsController.get(id);
     }
 
-    createTopic(id: TopicId, contentType: ContentType) {
```
ErykSol, line 56: I would move this arguments structure `(id: TopicId, contentType: ContentType)`, to `createTopicIfNotExist` for readibility and type safety, looks better than "magic" structure as arguments.

MichalCz: Can do.

packages/host/src/lib/serviceDiscovery/topic.ts
MichalCz: Not really - this can consume multiple piped streams one after another - it's a cleaner way of consuming the whole stream, and the `on("end")` is actually on the pipe, not on the `topic`.

a-tylenda:
### This PR additionally fixes two issues:
- https://github.com/scramjetorg/transform-hub/issues/914
- https://github.com/scramjetorg/transform-hub/issues/915

### but:
- There is no Error returned (`415`) when the topic is sent with the `application/x-ndjson` content type and collected with the `text/plain` content type:

image

Maybe it would not be a big deal, but the error is returned when you try another way around, which means, 
1. First `si topic get pets` (default content type `text/plain` is used)
2. Second `si topic send pets -t application/x-ndjson` (overwritten content type)

image

a-tylenda:
Also, the method `getNamedData()` has changed so that now, when you use it on hostClient you need to pass more arguments than just the topic name as it used to be before. 

```js
hostClient.getNamedData(topic, {}, "text/plain")
```

otherwise, you will get a bad request error 400:

image

a-tylenda (approved):

alicja-gruzdz (approved):

PR #960, opened 2023-10-12
Type change in order to achieve coherent structure for displaying

gzukowski:
**What?**  
Added a new type for displaying sequences.

**Why?**  
Docker's config contains "config" and "container" in fields, which are not coherent with the structure of displaying.

ClickUp task:
https://app.clickup.com/t/24308805/VDM-1440

- [ ] All STH tests pass

### DIFF
--- a/packages/adapters/src/docker-instance-adapter.ts
+++ b/packages/adapters/src/docker-instance-adapter.ts
@@ -171,7 +171,7 @@ IComponent {
 
     // eslint-disable-next-line complexity
     async run(config: InstanceConfig, instancesServerPort: number, instanceId: string): Promise<ExitCode> {
-        if (!("container" in config)) { //config.type !== "docker"
+        if (!(config.type === "docker" && "container" in config)) {
             throw new Error("Docker instance adapter run with invalid runner config");
         }
 
--- a/packages/adapters/src/docker-instance-adapter.ts
+++ b/packages/adapters/src/docker-instance-adapter.ts
@@ -171,7 +171,7 @@ IComponent {
 
     // eslint-disable-next-line complexity
     async run(config: InstanceConfig, instancesServerPort: number, instanceId: string): Promise<ExitCode> {
-        if (config.type !== "docker") {
+        if (!("container" in config)) { //config.type !== "docker"
             throw new Error("Docker instance adapter run with invalid runner config");
         }
 
--- a/packages/types/src/runner-config.ts
+++ b/packages/types/src/runner-config.ts
@@ -50,6 +50,6 @@ export type SequenceConfigDisplay = CommonSequenceConfig & {
 }
 
 // eslint-disable-next-line max-len
-export type SequenceConfig = DockerSequenceConfig | SequenceConfigDisplay | ProcessSequenceConfig | KubernetesSequenceConfig;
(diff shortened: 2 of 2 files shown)


packages/types/src/runner-config.ts
```diff
@@ -45,6 +45,11 @@ export type KubernetesSequenceConfig = CommonSequenceConfig & {
     type: "kubernetes"
 }
 
-export type SequenceConfig = DockerSequenceConfig | ProcessSequenceConfig | KubernetesSequenceConfig
+export type SequenceConfigDisplay = CommonSequenceConfig & {
+    type: string
+}
+
+// eslint-disable-next-line max-len
+export type SequenceConfig = DockerSequenceConfig| SequenceConfigDisplay | ProcessSequenceConfig | KubernetesSequenceConfig;
```
patuwwy, line 53:
```suggestion
export type SequenceConfig = DockerSequenceConfig | SequenceConfigDisplay | ProcessSequenceConfig | KubernetesSequenceConfig;
```

packages/adapters/src/docker-instance-adapter.ts
```diff
@@ -169,7 +169,7 @@ IComponent {
 
     // eslint-disable-next-line complexity
     async run(config: InstanceConfig, instancesServerPort: number, instanceId: string): Promise<ExitCode> {
-        if (config.type !== "docker") {
+        if (!("container" in config)) { //config.type !== "docker"
```
patuwwy, line 172:
```suggestion
        if (!(config.type == "docker" && ("container" in config))) {
```

?

gzukowski: I changed it because the creation of new type,  just for displaying, was connected with the problem  of possible "docker" type without container and config, which are used in this func.

gzukowski: You are right, yours is more readable.

packages/types/src/runner-config.ts
```diff
@@ -45,6 +45,11 @@ export type KubernetesSequenceConfig = CommonSequenceConfig & {
     type: "kubernetes"
 }
 
-export type SequenceConfig = DockerSequenceConfig | ProcessSequenceConfig | KubernetesSequenceConfig
+export type SequenceConfigDisplay = CommonSequenceConfig & {
+    type: string
```
ErykSol, line 49: `type: string` already exist in `CommonSequenceConfig` so this new type doesn't define any new type... `SequenceConfigDisplay=CommonSequenceConfig`

ErykSol (changes requested):

packages/types/src/runner-config.ts
gzukowski: You are right, I missed the fact that types like dockerconfig, kubernetsconfig overwrites the "type" field with the const. My bad.

ErykSol (approved):

alicja-gruzdz (approved):

kociolekscramjet (approved):
frontend will work with these changes

alicja-gruzdz:
https://github.com/scramjetorg/scramjet-cpm-dev/pull/619

PR #966, opened 2023-10-23
[CLI] Add `si seq start <seq-id> --inst-id` to CLI command

a-tylenda:
This option was missing. 

**To verify**  
- start sth from source: `yarn start:dev`
- send any sequence
```
yarn start:dev:cli seq send packages/hello-input-out.tar.gz
```
- start sequence with `--inst-id` option
```
yarn start:dev:cli seq start - --inst-id "Supervisor-Instance-0000-11111111111"
```

**Review checks:**

- [x] All STH tests pass

### DIFF
--- a/packages/cli/src/lib/config/sessionConfig.ts
+++ b/packages/cli/src/lib/config/sessionConfig.ts
@@ -1,7 +1,6 @@
 import { ConfigFileDefault } from "@scramjet/utility";
 import { SessionConfigEntity } from "../../types";
 import { sessionId } from "../../utils/sessionId";
-import isUUID from "validator/lib/isUUID";
 import { sessionConfigFile } from "../paths";
 
 // Session configuration represents configuration used internally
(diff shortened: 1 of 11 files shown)


packages/cli/src/lib/config/sessionConfig.ts
```diff
@@ -42,16 +41,15 @@ export class SessionConfig extends ConfigFileDefault<SessionConfigEntity> {
...
-    protected validateEntry(key: string, value: any): boolean | null {
+    protected validateEntry(key: string): boolean | null {
         switch (key) {
             case "lastPackagePath":
             case "lastSpaceId":
             case "lastSequenceId":
             case "lastHubId":
+            case "lastInstanceId":
             case "sessionId":
                 return null;
```
patuwwy, line 52: check if lastInstanceId is "" or has required length

a-tylenda: @patuwwy  
Instance ID is checked [here](https://github.com/scramjetorg/transform-hub/pull/966/files#diff-49139fa83ab4d9b3443fcffc2158a1f67af14baa9622bf060cd5c92c32a3343bR930) during sequence start, as a matter of fact whole payload is being checked here. We use a method [isStartSequenceEndpointPayloadDTO() ](https://github.com/scramjetorg/transform-hub/pull/966/files#diff-8da181dfd902335d0ab6991c5ad6c0edc82ebdd58ec454ab5320cd0f61b8adb8R25) that `instanceId` checks length, type and if it's not `undefined`. 
Do we need to check it again in the session config?

alicja-gruzdz (approved):

patuwwy (approved):

Issue #961, opened 2023-10-12
CLI does not work for more than one user on the same OS


alicja-gruzdz:
**Describe the bug**
CLI can not be used by more than one user on the same OS. It is due to the fact that CLI attempts to read user permission-restricted session files from the shared CLI /tmp system directory.

**Steps to Reproduce**
1. Install CLI on two user account on the same OS.
2. Execute any CLI command as the first user.
3. Execute any CLI command as the second user.
4. You will receive an error about session file permissions and the CLI command will fail to execute.
![Screenshot from 2023-10-12 15-10-55](https://github.com/scramjetorg/transform-hub/assets/8177865/06618d8a-0e4e-40c5-8dc0-05dcc275d76e)

**Expected behavior**
CLI must work for more than one user account on the OS.

**Version (please complete the following information):**
 - STH version: any release  with CLI sessions support.
 - node version: doesn't apply, but was tested on 16.18.0.
 - os: certainly Linux, maybe others [to verify].

PR #970, opened 2023-10-26
[CLI] Remove session configs with write access

patuwwy:
**What?**  
Deleting only those configuration files to which we have write permission

If several users use `si`, configuration files may be created with different permissions (depending on the system settings - umask). If a user using `si` automatically creates a directory with the configuration for the current session, it may not be deleted by another user when the session ends.
I added a check to see if we have write access when filtering directories to delete

**Review checks:**

- [x] All STH tests pass


alicja-gruzdz:
excellent, thanks, Patryk.
LInking issue https://github.com/scramjetorg/transform-hub/issues/961

alicja-gruzdz (approved):

PR #977, opened 2023-11-01
Fix logging circular args

patuwwy:
**What?**  
Latest change in logger has introduced error when logging circular arguments.
This is fix

**Why?**  

**Usage:**

-
-
-

**Clickup Task:** 

**How it works:**

-
-
-

**Review checks:**

- [ ] All STH tests pass

### DIFF
--- a/packages/obj-logger/src/obj-logger.ts
+++ b/packages/obj-logger/src/obj-logger.ts
@@ -122,7 +122,11 @@ export class ObjLogger implements IObjectLogger {
         let paramsCopy;
 
         if (optionalParams.length) {
-            paramsCopy = JSON.parse(JSON.stringify(optionalParams));
+            try {
+                paramsCopy = JSON.parse(JSON.stringify(optionalParams));
+            } catch {
+                paramsCopy = JSON.parse(JSON.stringify(optionalParams, getCircularReplacer()));
+            }
         }
 
         if (typeof entry === "string") {


packages/obj-logger/src/obj-logger.ts
```diff
@@ -122,7 +122,11 @@ export class ObjLogger implements IObjectLogger {
         let paramsCopy;
 
         if (optionalParams.length) {
-            paramsCopy = JSON.parse(JSON.stringify(optionalParams));
+            try {
+                paramsCopy = JSON.parse(JSON.stringify(optionalParams));
```
ErykSol, line 126:
```suggestion
               paramsCopy = JSON.parse(JSON.stringify(optionalParams, getCircularReplacer()));
```

packages/obj-logger/src/obj-logger.ts
```diff
@@ -122,7 +122,11 @@ export class ObjLogger implements IObjectLogger {
...
+            } catch {
+                paramsCopy = JSON.parse(JSON.stringify(optionalParams, getCircularReplacer()));
```
ErykSol, line 128:
```suggestion
               throw new Error("Error parsing logger params.");
```

ErykSol (changes requested):
Just a proposition, as it's more verbose on catch block about the problem

alicja-gruzdz (approved):

packages/obj-logger/src/obj-logger.ts
patuwwy: ```getCircularReplacer``` is passed as the second argument to ```JSON.stringify``` only if parsing without it returns an error. This way, if the object does not contain circular references, parsing without ```getCircularReplacer``` will be faster.

patuwwy: objects that we log and have circular references are really rare cases. We do not log them at all in transform-hub. However, it may happen that they are logged in from the Instance level. For this reason we do not need to check for circular references on every log.

Issue #328, opened 2022-03-05
Events stream format


patuwwy:
**Describe the bug**
when getting data from event stream
```
(await instanceClient.getEventStream("msg")).pipe(process.stdout)
```

i am receiving data in format:
```
{ "eventName": "msg", "message": { "a": 1 } }
{ "eventName": "msg", "message": { "a": 2 } }
{ "eventName": "msg", "message": { "a": 3 } }
```

I am requesting events with name `msg` so i don't need event name in response. As for me this information is redundant and could be bigger than payload itself.

**Expected behavior**
Just event data
```
{ "a": 1 }
{ "a": 2 }
{ "a": 3 }
```
**Version (please complete the following information):**
 - STH version: 0.18.4
 - node version: N/A
 - os: N/A

PR #987, opened 2023-11-13
Shorten instance/<instanceID>/events/:name endpoint response

piotrek6641:
Make events data more readable with removal of redundant data in output
`http://0.0.0.0:8000/api/v1/instance/{instanceId}/events/{eventName}`

before:
```
{ "eventName": "msg", "message": { "a": 1 } }
{ "eventName": "msg", "message": { "a": 2 } }
{ "eventName": "msg", "message": { "a": 3 } }
```
now returns:
```
{ "a": 1 }
{ "a": 2 }
{ "a": 3 }
```
without message used to return:
```
{ "eventName": "msg"  }
{ "eventName": "msg"  }
{ "eventName": "msg"  }
```
now returns 
```
 {}
 {}
 {}
```

### Screenshots
events with no message:
image

events with object as a message:
image

events with string as a message:
image

**Clickup Task:** 
https://app.clickup.com/t/24308805/VDM-1283
**Issue:**
https://github.com/scramjetorg/transform-hub/issues/328

### DIFF
--- a/packages/host/src/lib/csi-controller.ts
+++ b/packages/host/src/lib/csi-controller.ts
@@ -672,7 +672,14 @@ export class CSIController extends TypedEmitter<Events> {
             }
 
             const out = new DataStream();
-            const handler = (data: any) => out.write(data);
+            const handler = (data: any) => {
+                if (typeof data !== "object") {
+                    out.write(data);
+                }
+                const { eventName, ...message} = data;
+
+                out.write(message ? message : {} );
+            }
             const clean = () => {
                 this.logger.debug(`Event stream "${name}" disconnected`);
 
--- a/packages/host/src/lib/csi-controller.ts
+++ b/packages/host/src/lib/csi-controller.ts
@@ -662,9 +662,10 @@ export class CSIController extends TypedEmitter<Events> {
 
             if (!event.eventName) return;
 
-            localEmitter.lastEvents[event.eventName] = event;
+            localEmitter.lastEvents[event.eventName] = event.message;
             localEmitter.emit(event.eventName, event);
         });
+
         this.router.upstream("/events/:name", async (req: ParsedMessage, res: ServerResponse) => {
             const name = req.params?.name;
 
@@ -700,14 +701,14 @@ export class CSIController extends TypedEmitter<Events> {
             return out.JSONStringify();
         });
 
-        const awaitEvent = async (req: ParsedMessage): Promise<unknown> => new Promise(res => {
+        const awaitEvent = async (req: ParsedMessage): Promise<unknown> => new Promise((res) => {
             const name = req.params?.name;
 
             if (!name) {
                 throw new HostError("EVENT_NAME_MISSING");
             }
 
-            localEmitter.once(name, res);
+            localEmitter.once(name, (data) => res(data.message));
         });
 
         this.router.get("/event/:name", async (req) => {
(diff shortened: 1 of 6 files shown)


packages/types/src/api-expose.ts
```diff
@@ -6,7 +6,10 @@ import { ControlMessageCode, MonitoringMessageCode } from "./message-streams";
 import { IObjectLogger } from "./object-logger";
 import { MaybePromise } from "./utils";
 
-export type ParsedMessage = IncomingMessage & { body?: any; params?: { [key: string]: any } | undefined };
+export type ParsedMessage = IncomingMessage & {
+    body?: any;
+    params?: { [key: string]: any },
+    query?: { [key: string]: any }};
```
patuwwy, line 12:
```suggestion
    query?: { [key: string]: any }
};
```

ErykSol (approved):

packages/host/src/lib/csi-controller.ts
```diff
@@ -672,7 +672,15 @@ export class CSIController extends TypedEmitter<Events> {
             }
 
             const out = new DataStream();
-            const handler = (data: any) => out.write(data);
+            const handler = (data: any) => {
+                if (typeof data !== "object") {
+                    out.write(data);
```
patuwwy, line 677: missing return after write

patuwwy (changes requested):

packages/host/src/lib/csi-controller.ts
```diff
@@ -672,7 +672,15 @@ export class CSIController extends TypedEmitter<Events> {
...
+                }
+                // eslint-disable-next-line @typescript-eslint/no-unused-vars
+                const { _eventName, ...message } = data;
```
patuwwy, line 680: there is no `_eventName` in event message payload so this assignment will cause `message` to be `data`.

packages/host/src/lib/csi-controller.ts
```diff
@@ -672,7 +672,15 @@ export class CSIController extends TypedEmitter<Events> {
...
+
+                out.write(message ? message : {});
```
patuwwy, line 682: pls return just message from data or undefined if message is empty

packages/host/src/lib/csi-controller.ts
```diff
@@ -672,7 +672,15 @@ export class CSIController extends TypedEmitter<Events> {
...
+            };
```
patuwwy, line 683:
```suggestion
            const handler = (data: any) => {
                if (typeof data !== "object") {
                    out.write(data);
                    return;
                }

                out.write(data.message);
            };
```

patuwwy (approved):

patuwwy (approved):

a-tylenda:
I verified and endpoint `/api/v1/instance/{instanceId}/events/{eventName}` works as described in PR description, but...
there is a discrepancy between API and CLI. This is because under CLI command `si inst event on <id>` there is `getEvent()` method [this.router.get](https://github.com/scramjetorg/transform-hub/blob/feat/events-stream-format/packages/host/src/lib/csi-controller.ts#L713) not [this.router.upstream](https://github.com/scramjetorg/transform-hub/blob/feat/events-stream-format/packages/host/src/lib/csi-controller.ts#L668) edited by @piotrek6641

PR #993, opened 2023-11-21
[HOST] Api remove support for getSequenceByName

piotrek6641:
**What?**  
- Change host api Api sequence endpoint to be compatible with REST standard.
- Small changes to implement naming consistency 
- fix issues related to updating sequence via CLI. (When updating sequence cli was requesting using post method instead of put)

### Endpoints before:
`http://{host-apiBase}/sequence/:id_name`
Accepts both id/name

### After changes
`http://{host-apiBase}/sequence/:id`
Accepts only id now

###  New endpoint work structure works for all available methods for this url (GET, UPDATE, DELETE)

**Clickup Task:** 
https://app.clickup.com/t/24308805/VDM-1423

### DIFF
--- a/packages/host/src/lib/host.ts
+++ b/packages/host/src/lib/host.ts
@@ -781,15 +781,13 @@ export class Host implements IComponent {
             const existingSequence = this.sequenceStore.getById(id as string);
 
             if (existingSequence) {
-                if (req.method !== "put") {
-                    this.logger.debug("Overriding sequence", id, existingSequence.id);
-
+                if (req.method?.toLowerCase() !== "put") {
                     return {
                         opStatus: ReasonPhrases.METHOD_NOT_ALLOWED,
                         error: `Sequence with name ${id} already exist`
                     };
                 }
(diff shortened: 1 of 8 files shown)


packages/host/src/lib/host.ts
```diff
@@ -628,22 +628,23 @@ export class Host implements IComponent {
      * @returns {Promise<STHRestAPI.DeleteSequenceResponse>} Promise resolving to operation result object.
      */
     async handleDeleteSequence(req: ParsedMessage): Promise<OpResponse<STHRestAPI.DeleteSequenceResponse>> {
-        const id = req.params?.id;
+        if (!req.params?.id) return { opStatus: ReasonPhrases.BAD_REQUEST, error: "Missing id parameter" };
```
patuwwy, line 631:
```suggestion
        if (!req.params?.id) return { opStatus: ReasonPhrases.BAD_REQUEST, error: "Missing id parameter" };

```

PR #1000, opened 2023-12-01
VDM-1282: CLI autocomplete files and dirs

ErykSol:
-Removes outside package for completer to remove nesting, and replace it with new implementation extending features
-Adds autocomplete for files and dirs
-Adds autocomplete for option and argument choices params
-Adds event to handle more personalised values for arguments and options in future (instanceId etc)
-Fix install and uninstall script to work with other package managers and nvm
-Removes si completer bash option (not needed)
-Adds system check to allow `si completer` option only on linux OS
-Changes location of scripts from `cli/src/completion/` to `cli/scripts/completion`

Explanation of working on example:
i.e. `si inst input <id> [file] --content-type`
1) if command have children show chidren: i.e si or si inst - will show subcommands
2) if no more children check required arguments (no hihnts if not specified) i.e si inst input -will wait for user to write required id and no completer hinst should be given
3) if argument is optional give arg detailed hint (no hinsts if no event details provided: i.e si inst input 123 -will provide file hints
4) if no more arguments are required and command starts with -- provide options hint or options value hints if specified

**Setup for testing**

(requires global install of CLI)
1. In sth root: `yarn build:packages`
2. Get full path to sth/dist/cli/bin/index.js for later.
3. `which si` -> `<siFilePath>`
4. go to folder conatining `<siFilePath>` (without index.js)
5. In folder from 4. save value returned by `readlink -f si`
6.  rm si
7. `ln -s `<Path to dist/cli/bin/index.js file in point 2>` si` 
After all repeat 6. with file name from 5. to return to earlier version setup.

**How to TEST**
Run:
1. `si completion install` (just open new console or run `exec bash` to restart bash)
2. play with completer (si ... double TAB at variuos places)
3. `si completion uninstall`
2. play with completer (Now it should't work)

**Clickup Task:** 
https://app.clickup.com/t/24308805/VDM-1282

- [x] All STH tests pass
- [x] All [Scramjet Cloud Platform](https://docs.scramjet.org/platform) tests pass


alicja-gruzdz:
Verification of autocompletion for directories/filenames:
Screenshot from 2024-01-10 13-31-28
Screenshot from 2024-01-10 13-31-45

alicja-gruzdz:
Verification of autocompletion of `si` commands:
Screenshot from 2024-01-10 13-34-03

a-tylenda (approved):

PR #1001, opened 2023-12-01
Immediately send loadCheck to cmp when available

patuwwy:
**What?**  
Move call to `setLoadCheckMessageSender` after communicationStream is set.
Now loadCheck is called in second iteration because cs is not ready at first call

**Review checks:**

- [ ] All STH tests pass


a-tylenda (approved):
Great! thanks @patuwwy, this will speed up testing!

PR #1007, opened 2023-12-15
Faster eslint operation

MichalCz:
**What?**  

Runs eslint in modules so it doesn't need to load all the code.

**Why?**  

Eslint is failing a lot!

**Usage:**

No change.

**How it works:**

See change in package.json

**Review checks:**

- [ ] All STH tests pass


alicja-gruzdz (approved):
nice, thanks for speeding it up

patuwwy (approved):

PR #1026, opened 2024-02-05
move PANG message with topics after await result

RayNawfal:
**What?**  
fix specifying Topics on output and input stream and support async run in Topics 

**Usage:**
TBC

**Clickup Task:** 

**How it works:**

-
-
-

**Review checks:**

- [ ] All STH tests pass


alicja-gruzdz:
It solves the problem with Py Runner reporting topics to STH, but there is still an issue with data transfer in topics.

PR #1038, opened 2024-02-12
Remove docs for manager-api-client

alicja-gruzdz:
**What?**  
Remove manager-api-client docs

**Why?**  
The package was removed (Incorporated into api-client)


alicja-gruzdz:
thanks for spotting it @a-tylenda

a-tylenda (approved):

PR #1039, opened 2024-02-13
fix HTTPSAgent

karoltylenda:
**What?**  

Change HTTPAgent to HTTPSAgent

**Review checks:**

- [ ] All STH tests pass


patuwwy (changes requested):
Will not work with env development in cli

PR #1055, opened 2024-03-06
cli topic create connection

gzukowski:
**What?**  

Added handling the new topic from cli.

**Why?**  
After "si topic create", new topic was not visible in middleware.

**Review checks:**

- [ ] All STH tests pass

### DIFF
--- a/packages/api-client/src/host-client.ts
+++ b/packages/api-client/src/host-client.ts
@@ -228,7 +228,10 @@ export class HostClient implements ClientProvider {
         return this.client.getStream(`topic/${topic}`, requestInit, { type: contentType });
     }
 
-    async createTopic(id: string, contentType: string): Promise<{ topicName: string }> {
+    async createTopic(id: string, contentType: string, cli? : boolean): Promise<{ topicName: string }> {
+        if (cli) {
+            return this.client.post("topics", { id, "content-type": contentType, cli }, undefined, { json: true, parse: "json" });
+        }
         return this.client.post("topics", { id, "content-type": contentType }, undefined, { json: true, parse: "json" });
     }
 
--- a/packages/api-client/src/host-client.ts
+++ b/packages/api-client/src/host-client.ts
@@ -228,10 +228,7 @@ export class HostClient implements ClientProvider {
         return this.client.getStream(`topic/${topic}`, requestInit, { type: contentType });
     }
 
-    async createTopic(id: string, contentType: string, cli? : boolean): Promise<{ topicName: string }> {
-        if (cli) {
-            return this.client.post("topics", { id, "content-type": contentType, cli }, undefined, { json: true, parse: "json" });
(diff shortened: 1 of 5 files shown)


packages/api-client/src/host-client.ts
```diff
@@ -228,7 +228,10 @@ export class HostClient implements ClientProvider {
         return this.client.getStream(`topic/${topic}`, requestInit, { type: contentType });
     }
 
-    async createTopic(id: string, contentType: string): Promise<{ topicName: string }> {
+    async createTopic(id: string, contentType: string, cli? : boolean): Promise<{ topicName: string }> {
+        if (cli) {
+            return this.client.post("topics", { id, "content-type": contentType, cli }, undefined, { json: true, parse: "json" });
+        }
```
patuwwy, line 234: This can't be done like this. 
Host client is a library which doesn't and can't know about CLI.

packages/host/src/lib/serviceDiscovery/sd-adapter.ts
```diff
@@ -94,13 +94,24 @@ export class ServiceDiscovery {
...
             this.logger.debug("Topic routed:", config);
+
             return topic;
         }
 
         this.logger.debug("Topic added:", config);
         const origin: StreamOrigin = { id: this.hostName, type: "hub" };
         const newTopic = new Topic(topicName, config.contentType, origin);
 
+        if (cli) {
```
patuwwy, line 105: send it always, sd discover can't know about CLI

patuwwy (changes requested):

packages/host/src/lib/serviceDiscovery/sd-adapter.ts
gzukowski: createTopicIfNotExist is already mentioned in two other methods which already sends the info to cpm, am I supposed to move these to the createTopicIfNotExist ?

patuwwy: do not mix responsibility - createTopicIfNotExist creates topic. 
Create a method to send topicInfo or call sendTopicInfo directly after topic is created - after createTopicIfNotExist is called

packages/host/src/lib/serviceDiscovery/sd-adapter.ts
```diff
@@ -79,7 +79,7 @@ export class ServiceDiscovery {
      * Returns topic with given configuration, if not exists creates new one.
      *
      * @param {DataType} config Topic configuration.
-     * @param {string} [localProvider] Provider identifier. It not set topic will be considered as external.
+     * @param {string} cli Flag specifying the origin of the request from cli.
```
patuwwy, line 82:
```suggestion
```

patuwwy (approved):

PR #1068, opened 2024-03-20
Fix reconnect logging and memleak

MichalCz:
**What?**  

* Fix the logging levels on reconnect from the runner
* Unhook streams before reconnecting an instance

**Why?**  

Currently a reconnect causes a double call to `hookupStreams` which pipes multiple streams to the input. This unhooks the streams on reconnect, so there are no extra listeners and we're not causing a refleak.

This only happens when a sequence stream errors out.

Additionally stream connect and reconnect errors are muted while connecting and logging happens only on disconnection.

**Reproduce:**

- Create an instance that sends an error on output (`out.emit("error", new Error())`)
- Observe the chain of reconnects

**Review checks:**

- [ ] All STH tests pass

### DIFF
--- a/packages/host/src/lib/csi-controller.ts
+++ b/packages/host/src/lib/csi-controller.ts
@@ -169,7 +169,7 @@ export class CSIController extends TypedEmitter<Events> {
     /**
      * Streams connected do API.
      */
-    private downStreams?: DownstreamStreamsConfig;
+    private downStreams: DownstreamStreamsConfig | null = null;
     private upStreams: PassThroughStreamsConfig;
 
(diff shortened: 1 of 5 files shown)


MichalCz:
> We need to double check if an error in sequence does kill it.

EDIT: fixed it.

patuwwy (approved):

alicja-gruzdz:
It does not break the existing functionalities of Reconnect.
Side note: it did not fix https://app.clickup.com/t/86944kkf4

alicja-gruzdz (approved):

